PR #25, opened 2016-03-24
Code quality fix - Try-with-resources should be used.

faisal-hameed:
This pull request is focused on resolving occurrences of Sonar rule 
squid:S2093 -  Try-with-resources should be used.
You can find more information about the issue here: 
https://dev.eclipse.org/sonar/rules/show/squid:S2093

Please let me know if you have any questions.

Faisal Hameed


iocanel:
the compiler plugin is configured with source 1.5 and target 1.5 which means that its not possible to use try with resources etc.

Issue #43, opened 2016-06-01
Fix plurals


jimmidyson:
Properties like `addresses` have fluent methods created like `addToAddresse` (notice the extra e). Should be able to support most standard plurals & perhaps make it configurable what the singular/plural form should be.

PR #45, opened 2016-06-07
#43: Fixed singularize.

iocanel:


jimmidyson:
Where did you get that word list from?

iocanel:
There is a link in the source.

PR #48, opened 2016-06-10
Handle missing getter info

jimmidyson:


iocanel:
Find getters itself is working fine!

There are two issues that could possibly affect it

1. buildable references are broken potentially breaking isAssignable() -> findGetter().
2. buildable constructor detection is also broken. This means that sundrio may pick up the wrong constructor and start looking for fields that are not supposed to have getters.

jimmidyson:
I think is a safe defensive test if there is nothing to iterate over so...

iocanel:
ok I am blind! Didn't realize this was a PR! LoL

PR #66, opened 2016-09-23
Adding support for external imported BOMs

nicolaferraro:
I added the support for importing external BOMs directly in the main mojo. I added a `imports` child into the configuration, so that a BOM can be composed of direct dependencies of the children modules plus external managed dependencies.

I also added a check to verify if the dependencies of the generated BOMs have conflicts, i.e. same library with different version (this is the case for some BOMs generated in fabric8, where eg. there are 3-4 different versions of `plexus-utils` in the same BOM, the first one wins).
By default a warning is printed to the log, but a flag can turn this to a build failure.

I added support for exclusions in dependency management, because external BOMs usually force you to exclude some transitive dependencies when using a managed lib (eg. with spring-boot managed dependencies you need to exclude all transitive logging dependencies, to prevent conflicts with the embedded slf4j logging mechanism).

### DIFF
--- a/maven-plugin/pom.xml
+++ b/maven-plugin/pom.xml
@@ -62,6 +62,17 @@
                     </execution>
                 </executions>
             </plugin>
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-plugin-plugin</artifactId>
+                <version>${maven-plugin-plugin-version}</version>
+                <executions>
+                    <execution>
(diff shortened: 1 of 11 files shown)


iocanel:
Sorry for being slow on reviewing this. I will look at it asap.

nicolaferraro:
@iocanel thanks, np :)

iocanel:
I love the fact that it fits the existing mojo like a glove :-)

PR #72, opened 2016-12-07
Attributes

iocanel:


annotations/builder/src/main/resources/snippets/validate.txt
```diff
@@ -0,0 +1,25 @@
...
+  StringBuilder sb = new StringBuilder("Constraint Validations: ");
+  boolean first = true;
+  for (ConstraintViolation violation: violations) {
+     if (first) {
+        first = false;
+     } else {
+       sb.append(", ");
+     }
+     Object leafBean = violation.getLeafBean();
+     sb.append(violation.getPropertyPath() + " " + violation.getMessage() + " on bean: " + leafBean);
```
jimmidyson: Does it make sense to format with new lines?

iocanel: Its a matter of taste really!

I can see pros and cons either ways.

iocanel: @jstrachan: wdyt? I will release later tonight!

jstrachan: thanks - looks good. I figured the generation of the message might just be a helper function; to keep the code generation simpler (and result in less bytecode)? i.e. just have a static function which generates a canonical message given an `Iterable< ConstraintViolation>`

iocanel: Let me see if this can move inside BaseFluent as a static method.

PR #75, opened 2016-12-19
Allow overriding the version of specific artifacts in the generated BOM

nicolaferraro:
This allows forcing the version of specific dependencies in a BOM when the one shipped with eg. an external BOM is not good anymore. This is useful in the cases where the external BOM cannot be changed, probably because it belongs to others..

@jamesnetherton maybe you want to check if this is useful for our needs (eg. for case of slf4j).

### DIFF
--- a/maven-plugin/src/it/multimodule/pom.xml
+++ b/maven-plugin/src/it/multimodule/pom.xml
@@ -218,6 +218,50 @@
                                 </excludes>
                             </goals>
                         </bom>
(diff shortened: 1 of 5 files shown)


jamesnetherton:
@iocanel Any thoughts on this? Would be useful to get this merged (and released) if possible.

iocanel:
@nicolaferraro @jamesnetherton: I'll release soon.

PR #84, opened 2017-11-29
feat: Add java.util.Optional support

jimmidyson:


annotations/builder/src/main/java/io/sundr/builder/internal/functions/ClazzAs.java
```diff
@@ -92,8 +89,11 @@ public TypeDef apply(TypeDef item) {
                 boolean isSet = TypeUtils.isSet(toAdd.getTypeRef());
                 boolean isList = TypeUtils.isList(toAdd.getTypeRef());
                 boolean isMap = TypeUtils.isMap(toAdd.getTypeRef());
-                boolean isCollection = isSet || isList;
                 boolean isAbstract = isAbstract(unwrapped);
+                boolean isOptionalInt = TypeUtils.isOptionalInt(toAdd.getTypeRef());
```
iocanel: Why do we care about the generic type? 
It seems to me that we just need to know if type is Optional, no?

annotations/builder/src/main/java/io/sundr/builder/internal/functions/ClazzAs.java
```diff
@@ -214,12 +217,15 @@ public TypeDef apply(TypeDef item) {
                 final boolean isSet = TypeUtils.isSet(property.getTypeRef());
                 final boolean isList = TypeUtils.isList(property.getTypeRef());
                 final boolean isMap = TypeUtils.isMap(property.getTypeRef());
-                final boolean isCollection = isSet || isList;
                 final boolean isAbstract = isAbstract(unwrapped);
+                final boolean isOptionalInt = TypeUtils.isOptionalInt(property.getTypeRef());
```
iocanel: Same. It seem that we just want to know if its optional.

annotations/builder/src/main/java/io/sundr/builder/internal/functions/ClazzAs.java
jimmidyson: `OptionalInt` is not just an optional... They require different handling: `OptionalInt.empty()`, `OptionalInt.of(1)`, etc. See InitEnricher and BuilderUtils (although I'm not sure those init statements are used anywhere any more).

iocanel: Sure, but we don't seem to handle it here.

annotations/builder/src/main/java/io/sundr/builder/internal/functions/ToMethod.java
```diff
@@ -137,17 +144,21 @@ public Method apply(Property property) {
                 } else if (IS_LIST.apply(type) || IS_SET.apply(type)) {
                     statements.add(new StringStatement("if (this." + fieldName + " == null) { this." + fieldName + " = " + property.getAttribute(LAZY_INIT) + ";} else {_visitables.removeAll(this."+fieldName+"); this." + fieldName + ".clear();}"));
                     String addToMethodName = "addTo" + property.getNameCapitalized();
-                    statements.add(new StringStatement("if (" + argumentName + " != null) {for (" + unwraped.toString() + " item : " + argumentName + "){this." + addToMethodName + "(item);}} return (" + returnType + ") this;"));
+                    statements.add(new StringStatement("if (" + argumentName + " != null) {for (" + unwrapped.toString() + " item : " + argumentName + "){this." + addToMethodName + "(item);}} return (" + returnType + ") this;"));
                 }
                 return statements;
-            } else if (isBuildable(unwraped) && !isAbstract(unwraped)) {
```
iocanel: Did u hit any issues and u decided to turn `else if ` into a plain `if`?
How would this code work, if u had to deal with a `List` of `Buildable` items?

annotations/builder/src/main/java/io/sundr/builder/internal/functions/ToMethod.java
```diff
@@ -137,17 +144,21 @@ public Method apply(Property property) {
...
-                TypeDef builder = BUILDER.apply(((ClassRef) unwraped).getDefinition());
+            }
+
+            if (isBuildable(unwrapped) && !isAbstract(unwrapped)) {
+                TypeDef builder = BUILDER.apply(((ClassRef) unwrapped).getDefinition());
                 String builderClass = builder.toReference().getName();
                 statements.add(new StringStatement("if (" + argumentName + "!=null){ this." + fieldName + "= new " + builderClass + "(" + argumentName + "); _visitables.add(this." + fieldName + ");} return (" + returnType + ") this;"));
                 return statements;
-            } else if (!descendants.isEmpty()) {
```
iocanel: Same as above.

annotations/builder/src/main/java/io/sundr/builder/internal/functions/ToMethod.java
jimmidyson: Each if else block had a return statement in so the else bit is redundant - will return or drop through. Just made it more readable, no problems.

jimmidyson: What I mean is:

```java
if (...) {
     return ...;
} else if (...) {
     return ...;
} else if (...) {
     return ...;
} else {
     return ...;
}
```

is the same functionally without the redundant elses as:

```java
if (...) {
     return ...;
}
if (...) {
     return ...;
}
if (...) {
     return ...;
}

return ...;
```

annotations/builder/src/main/java/io/sundr/builder/internal/functions/ClazzAs.java
jimmidyson: Ah OK - will refactor that out.

annotations/builder/src/main/java/io/sundr/builder/internal/functions/ClazzAs.java
jimmidyson: Updated.

annotations/builder/src/main/java/io/sundr/builder/internal/functions/ClazzAs.java
jimmidyson: Updated.

jimmidyson:
So how best to handle building with Java 7? Perhaps build with Java 8 but with Java 7 compatibility?

iocanel:
You can't have both! You need to have one or the other!

Until now, it used to require jdk7 for compilation, but sundrio itslef worked with all java version (even 5).

annotations/builder/src/main/java/io/sundr/builder/internal/functions/ToMethod.java
iocanel: Right! Didn't realize there was a return statement.

annotations/builder/src/main/java/io/sundr/builder/internal/utils/BuilderUtils.java
```diff
@@ -423,7 +425,7 @@ public static String fullyQualifiedNameDiff(TypeRef typeRef, TypeDef originType)
         String currentPackage = originType != null ? originType.getPackageName() : null;
 
         if (typeRef instanceof ClassRef) {
-            ClassRef classRef = (ClassRef)  TypeAs.combine(UNWRAP_COLLECTION_OF, UNWRAP_ARRAY_OF).apply(typeRef);
+            ClassRef classRef = (ClassRef)  TypeAs.combine(UNWRAP_COLLECTION_OF, UNWRAP_ARRAY_OF, UNWRAP_OPTIONAL_OF).apply(typeRef);
```
iocanel: @jimmidyson: You can't just cast a TypeRef to ClassRef without checking. I know that the codebase is allover this kind of cast, but this is only the case, when we are dealing with types recognized as buildables. So we now that they are not primitves, they are not generics, or null types.

annotations/builder/src/main/java/io/sundr/builder/internal/utils/BuilderUtils.java
```diff
@@ -463,17 +465,48 @@ public static Property buildableField(Property property) {
...
                     .addToAttributes(ALSO_IMPORT,  alsoImport(property, setRef, builderType))
                     .build();
-        } else {
-            return new PropertyBuilder(property).withTypeRef(builderType)
+        }
+
+        if (TypeUtils.isOptionalLong(classRef)) {
+            ClassRef optionalRef = Optionals.OPTIONAL_LONG.toReference(builderType);
+            return new PropertyBuilder(property).withTypeRef(optionalRef)
+                .addToAttributes(INIT, " " + optionalRef + ".empty()")
```
iocanel: There is no value in using the actual ref (as it causes issues in many cases). Just use OptionalLong.empty(). Easier and safer.

PR #99, opened 2018-04-18
feat: Support an extraDependencies config option for manually adding dependencies to the boms.

chirino:


chirino:
@iocanel could you merge this and kick off a release?

Issue #129, opened 2019-02-21
Running builder with Java 11 fails


ctron:
Having a project which is targeted for Java 8, and running the Maven build with Java 11 fails the build with the following output:

~~~
…
[ERROR] /home/jreimann/gobase/enmasse/src/github.com/enmasseproject/enmasse/api-model/target/generated-sources/annotations/io/enmasse/user/model/v1/UserSpecFluent.java:[6,17] java.lang.AbstractStringBuilder is not public in java.lang; cannot be accessed from outside package
[ERROR] /home/jreimann/gobase/enmasse/src/github.com/enmasseproject/enmasse/api-model/target/generated-sources/annotations/io/enmasse/user/model/v1/UserSpecFluent.java:[26,30] java.lang.AbstractStringBuilder is not public in java.lang; cannot be accessed from outside package
[ERROR] /home/jreimann/gobase/enmasse/src/github.com/enmasseproject/enmasse/api-model/target/generated-sources/annotations/io/enmasse/user/model/v1/UserAuthenticationFluent.java:[7,17] java.lang.AbstractStringBuilder is not public in java.lang; cannot be accessed from outside package
…
~~~

The same build works fine when running the Maven build with Java 8.

ctron:
In the generated code you can actually see an import for that:

~~~java
import java.lang.AbstractStringBuilder;
~~~

<details>
<summary>Full source</summary>

~~~java
package io.enmasse.address.model;

import com.fasterxml.jackson.annotation.JsonProperty;
import java.lang.StringBuilder;
import io.fabric8.kubernetes.api.builder.Nested;
import java.lang.AbstractStringBuilder;
import java.util.ArrayList;
import java.lang.String;
import io.fabric8.kubernetes.api.builder.Predicate;
import java.lang.Deprecated;
import com.fasterxml.jackson.annotation.JsonSetter;
import java.util.List;
import java.lang.Boolean;
import java.lang.Void;
import io.enmasse.admin.model.v1.NetworkPolicy;
import io.fabric8.kubernetes.api.builder.Fluent;
import io.enmasse.admin.model.v1.NetworkPolicyBuilder;
import java.lang.StringBuffer;
import javax.validation.Valid;
import io.enmasse.admin.model.v1.NetworkPolicyFluent;
import java.util.Collection;
import javax.validation.constraints.NotEmpty;

public interface AddressSpaceSpecFluent<A extends AddressSpaceSpecFluent<A>> extends Fluent<A>{


[...]

PR #131, opened 2019-02-21
[#129]: Fix invalid code, generated when running with Java 11

ctron:


iocanel:
@ctron: being a complete java 11 noob here. What exactly is that change that triggeres this issue? I looked at the java docs for java.reflect.Methos but they weren't that helpful.

ctron:
The problem was already there, as the lookup code, which looks for constructors is can call, doesn't take package level visibility under consideration. So you can run into this issue on Java 8 already.

However it seems that something in Java 11 has changed, and I don't know what exactly, but it reproducibly creates a reference to `AbstractStringBuilder`, which is wrong:

~~~
public A withNewName(AbstractStringBuilder arg1,Void arg2)
public A withNewName(String arg1)
public A withNewName(StringBuilder arg1)
public A withNewName(StringBuffer arg1)
~~~

The issue is fixed in any case if the filter checks for the constructor being either public or protected, which filters out private and "default" (package level).

iocanel:
@ctron: Thanks!

I'll merge it and try to have a release out later today!

iocanel:
@ctron: Release 0.16.4.

ctron:
> @ctron: Released 0.16.4.
Thanks!

PR #144, opened 2019-05-11
fix: do not call super.equals if super == Object.

metacosm:


metacosm:
This was an attempt to get a fix for https://github.com/snowdrop/istio-java-api/issues/66 but the `istio-java-api` builder objects actually are children of `io.fabric8.kubernetes.api.builder.BaseFluent` (as opposed to `io.sundr.builder.BaseFluent`) which doesn't have an `equals` implementation, so this fix doesn't actually solve the issue anyway…

metacosm:
For a proper fix, we might need to take the `@Buildable.builderPackage` value into account, somehow. Not sure how, though, at this point…

naitikdani:
@iocanel, when do you think this issue will be fixed?

PR #148, opened 2019-05-17
Improve handling of removeMatching* methods to better handle inheritance.

iocanel:


iocanel:
@metacosm If you could give it a review, we could merge and release.

Basically, it does two things:

1. change the name from removeAllMatching* to removeMatching as all feels a bit redundant.
2 Create a remove from matching for descendant classes (see in shapes example how we now generate removeMatchingCircleShapes etc.

PR #155, opened 2019-08-10
Support for maps with buildable values.

mgorniew:
This will add nested builder for map properties with buildable values.

```java
@Buildable
public class Example {
    private Map<String, Person> people;
    // setters, etc.
}

@Buildable
public class Person {
    private String firstName;
    private String lastName;
    // setters, etc.
}

new ExampleBuilder()
        .addNewValueToPeople("person1")
            .withFirstName("FirstName1").withLastName("LastName1").endPeople()
        .addNewValueLikeToPeople("person2", new Person("a", "b"))
            .withFirstName("FistName2").endPeople()
        .editValueInPeople("person2")
            .withLastName("LastName2").endPeople()
        .editOrAddValueInPeople("person3")
            .withFirstName("FirstName3").endPeople()
        .build();
```

### DIFF
--- a/annotations/builder/src/main/java/io/sundr/builder/internal/functions/ClazzAs.java
+++ b/annotations/builder/src/main/java/io/sundr/builder/internal/functions/ClazzAs.java
@@ -36,6 +36,7 @@ import io.sundr.codegen.model.AnnotationRefBuilder;
 import io.sundr.codegen.model.Block;
 import io.sundr.codegen.model.ClassRef;
 import io.sundr.codegen.model.ClassRefBuilder;
+import io.sundr.codegen.model.EditableClassRef;
 import io.sundr.codegen.model.Method;
 import io.sundr.codegen.model.MethodBuilder;
 import io.sundr.codegen.model.Property;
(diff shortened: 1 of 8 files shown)


iocanel:
That's a great idea! I always wanted to support something like this but never found the time to dive into it. I will have a look soon.

iocanel:
@mgorniew: Thanks again for the pull request and apologies for the late response. Let's work on those details so that we can merge.

mgorniew:
Issue was caused by caching ADD_NEW_VALUE_TO_MAP function. I've fixed this and now kubernetes client model will compile, but their test will still fail. This is caused by [change](https://github.com/sundrio/sundrio/commit/3f8bf09e90e320868ec16bd098810174ca7de60a#diff-b0dcabfb7893ab6b6006799641f220efR145) in default map initialization.

iocanel:
@mgorniew: That's awesome! I'll try to have a look later tonight.

iocanel:
@mgorniew: do you intend to add tests?

mgorniew:
Yes. I will add some tests on weekend.

czw., 7 lis 2019, 11:37 użytkownik Ioannis Canellos <
notifications@github.com> napisał:

iocanel:
@mgorniew: Thanks! Give me a ping when you do!
BTW, it doesn't have to be necessarily in the form of a traditional unit test. 
An example accompanied with a test case that ensures the generated code will do.

mgorniew:
I've added support for abstract types and some tests in shapes examples project.

PR #158, opened 2019-11-20
fix: Handle consistnently name diffs when there are name clashes.

iocanel:


iocanel:
@metacosm: can you please let me know if this fixes the issue for you?

metacosm:
Yep, going to give a try now. Is it normal that CI fails, though?

iocanel:
@metacosm: Ooops didn't realize that. Let me fix that and you can try later.

metacosm:
It indeed doesn't build locally either…

Issue #160, opened 2019-12-13
java.lang.OutOfMemoryError: GC overhead limit exceeded  during Builds during annotation processing


rohanKanojia:
Hi,

We are using this project in Fabric8 Kubernetes Client for the generation of helper classes. Recently we have been facing lots of `java.lang.OutOfMemoryError: GC overhead limit exceeded` in our CI during the annotation processing step. Here are some logs:

```
98%: io.fabric8.kubernetes.api.model.CSIPersistentVolumeSource
Generating: io.fabric8.kubernetes.api.model.CSIPersistentVolumeSourceFluent
Generating: io.fabric8.kubernetes.api.model.CSIPersistentVolumeSourceFluentImpl
Generating: io.fabric8.kubernetes.api.model.CSIPersistentVolumeSourceBuilder
Generating: io.fabric8.kubernetes.api.model.DoneableCSIPersistentVolumeSource
98%: io.fabric8.kubernetes.api.model.PersistentVolumeClaimList
Generating: io.fabric8.kubernetes.api.model.PersistentVolumeClaimListFluent
Generating: io.fabric8.kubernetes.api.model.PersistentVolumeClaimListFluentImpl
Generating: io.fabric8.kubernetes.api.model.PersistentVolumeClaimListBuilder
Generating: io.fabric8.kubernetes.api.model.DoneablePersistentVolumeClaimList
98%: io.fabric8.kubernetes.api.model.certificates.CertificateSigningRequest
Generating: io.fabric8.kubernetes.api.model.certificates.CertificateSigningRequestFluent
Generating: io.fabric8.kubernetes.api.model.certificates.CertificateSigningRequestFluentImpl
Generating: io.fabric8.kubernetes.api.model.certificates.CertificateSigningRequestBuilder
Generating: io.fabric8.kubernetes.api.model.certificates.DoneableCertificateSigningRequest
99%: io.fabric8.kubernetes.api.model.ConfigMapVolumeSource
Generating: io.fabric8.kubernetes.api.model.ConfigMapVolumeSourceFluent
Generating: io.fabric8.kubernetes.api.model.ConfigMapVolumeSourceFluentImpl
Generating: io.fabric8.kubernetes.api.model.ConfigMapVolumeSourceBuilder
Generating: io.fabric8.kubernetes.api.model.DoneableConfigMapVolumeSource
99%: io.fabric8.kubernetes.api.model.rbac.RoleBindingList
[...]

PR #161, opened 2019-12-19
#160 refactor: Memory efficient ClassRef instances (reuse)

manusa:
Optimizes codegen memory usage.

Use `HashMap` (less memory footprint) instead of `LinkedHashMap` where applicable.

Cache `EditableClassRef` [instances](https://en.wikipedia.org/wiki/Flyweight_pattern).

HeapDump analysis of Kubernetes-client compilation shows an enormous amount of `EditableClassRef` instances being generated:
201912171040NOOptimizations

Same HeapDump after refactor implementation cuts in half the number of EditableClassRef, considerably reducing the amount of memory consumed by the process:
201912191252-ClassRefCache

Improves: #160

### DIFF
--- a/codegen/src/main/java/io/sundr/codegen/model/AnnotationRefFluentImpl.java
+++ b/codegen/src/main/java/io/sundr/codegen/model/AnnotationRefFluentImpl.java
@@ -18,23 +18,23 @@ package io.sundr.codegen.model;
 
 import io.sundr.builder.Nested;
 
-import java.util.LinkedHashMap;
+import java.util.HashMap;
 import java.util.Map;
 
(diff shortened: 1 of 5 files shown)


rohanKanojia:
@manusa @iocanel : I tested this and it seems to be working. With this version of sundrio, I was able to compile Fabric8 Kubernetes Client with `-xMx=2g`.

@iocanel : Would appreciate it if you could review it whenever you come back from holidays.

rohanKanojia:
@iocanel : Would it be possible for you to cut a release anytime soon? I was wondering if we could get this in before releasing Fabric8 Kubernetes Client 4.7.0.

iocanel:
I'll try to have one later today or tomorrow!

rohanKanojia:
Thanks a lot

PR #185, opened 2020-12-17
Typo

sknot-rh:
Signed-off-by: Stanislav Knot <sknot@redhat.com>

I believe there is a typo in the exception. See the message below.
```
java.lang.IllegalStateException: Cannot use different builder package names in a single project. Used:io.fabric8.kubernetes.api.builder but package:false already exists.
```
It prints `false` instead of the package name.


metacosm (approved):
LGTM

PR #195, opened 2021-02-23
fix: ClassTo now supports declaring class and generic properties

iocanel:


metacosm (approved):
Seems to address the issue I was seeing. Can't really comment on the code itself, just on its observed behavior… 😄

PR #201, opened 2021-03-12
Reduce output verbosity by reusing the progress line

jorsol:
The output when generating a large model can be annoying, this PR tries to reduce the verbosity and at the same time do some formatting of the output by reusing the lines.

Some "Generating:" lines make the output unnecessary verbose so they are commented, the only lines that are keep printed in a new line are the "Skipping:" (although they might be unnecessary too).

### DIFF
--- a/codegen/src/main/java/io/sundr/codegen/processor/JavaGeneratingProcessor.java
+++ b/codegen/src/main/java/io/sundr/codegen/processor/JavaGeneratingProcessor.java
@@ -20,9 +20,11 @@ import java.io.ByteArrayInputStream;
 import java.io.IOException;
 import java.io.StringWriter;
 import java.io.Writer;
+import java.nio.charset.StandardCharsets;
 
 import javax.annotation.processing.AbstractProcessor;
 import javax.annotation.processing.FilerException;
+import javax.tools.Diagnostic;
 import javax.tools.FileObject;
 import javax.tools.JavaFileObject;
 import javax.tools.StandardLocation;
@@ -40,7 +42,7 @@ public abstract class JavaGeneratingProcessor extends AbstractProcessor {
 
   /**
    * Generates a source file from the specified {@link io.sundr.codegen.model.TypeDef}.
-   * 
+   *
    * @param model The model of the class to generate.
    * @param resourceName The template to use.
    * @throws IOException If it fails to create the source file.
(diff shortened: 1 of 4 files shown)


codegen/src/main/java/io/sundr/codegen/processor/JavaGeneratingProcessor.java
```diff
@@ -65,10 +65,10 @@ public void generateFromResources(TypeDef model, String resourceName) throws IOE
    */
   public void generateFromResources(TypeDef model, JavaFileObject fileObject, String resourceName) throws IOException {
     if (classExists(model)) {
-      System.err.println("Skipping: " + model.getFullyQualifiedName() + ". Class already exists.");
+      printSkipping("Class", model.getFullyQualifiedName());
       return;
     }
-    System.err.println("Generating: " + model.getFullyQualifiedName());
+    // System.err.println(String.format("Generating: %-100s\r", model.getFullyQualifiedName()));
```
iocanel, line 71: Is this comment intended?

jorsol: Yes, is intended, it can be removed as this make the output unnecessarily verbose.

But it can be uncommented too if you think this is needed.

iocanel: Let's remove it completely!

jorsol: Done!

codegen/src/main/java/io/sundr/codegen/processor/JavaGeneratingProcessor.java
```diff
@@ -207,4 +207,9 @@ private static boolean classExists(TypeDef typeDef) {
       return false;
     }
   }
+
+  private void printSkipping(String name, String msg) {
+    processingEnv.getMessager().printMessage(Diagnostic.Kind.OTHER,
```
jorsol, line 212: @iocanel is the "Skipping" message a warning or info?

iocanel: Info or debug

iocanel (approved):
LGTM

PR #211, opened 2021-04-13
fix: ClassTo now corectlly processes interfaces and superclass

iocanel:


codegen/src/main/java/io/sundr/codegen/functions/ClassTo.java
```diff
@@ -142,8 +142,12 @@ public TypeDef apply(Class item) {
 
       if (item.getSuperclass() != null) {
         extendsList.add((ClassRef) TYPEREF.apply(item.getGenericSuperclass()));
+        references.add(item.getSuperclass());
```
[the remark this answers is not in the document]
iocanel: I think that it doesn't really matter as what we need is to generate the definition for the class itself, we don't really care about the arguments (as they are irrelevant to the definition).

codegen/src/main/java/io/sundr/codegen/functions/ClassTo.java
```diff
@@ -142,8 +142,12 @@ public TypeDef apply(Class item) {
...
       }
 
+      for (Class interfaceClass : item.getInterfaces()) {
```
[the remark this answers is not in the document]
iocanel: same

codegen/src/main/java/io/sundr/codegen/functions/ClassTo.java
```diff
@@ -191,6 +197,7 @@ public TypeDef apply(Class item) {
 
       Set<Class> copy = new HashSet<>(references);
       copy.stream()
+          .peek(c -> references.remove(c))
```
metacosm, line 200: Why is remove called here, when it's also called in `forEach`?

iocanel: This was meant to replace the remove in forEach, so that we always remove items, even filtered ones.

metacosm: It doesn't sound like a good idea to have side effects in a call to `peek`…

iocanel: why?

metacosm: See the `peek` javadoc :)

iocanel: Sure, but that doesn't mean that I have to duplicate the loop just because something is marked `mostly for logging`.

PR #215, opened 2021-04-19
fix: properties should have their associated annotation parameters

### LINKED ISSUE #214: ClassTo.ANNOTATIONTYPEREF doesn't provide annotations' parameters

metacosm:
Fixes #214 (sorta). Turns out that annotations are rather painful to
process as you cannot call a method that you got from the annotation
type directly on it, so a proper fix would be to pass the actual
annotation instead of the annotation type so that both are available
during the processing.

### DIFF
--- a/codegen/src/main/java/io/sundr/codegen/functions/ClassTo.java
+++ b/codegen/src/main/java/io/sundr/codegen/functions/ClassTo.java
@@ -261,11 +261,12 @@ public class ClassTo {
       AnnotationRef annotationRef = ANNOTATIONTYPEREF.apply(annotationType);
       Map<String, Object> parameters = new HashMap<>();
       for (java.lang.reflect.Method method : annotationType.getDeclaredMethods()) {
+        final String name = method.getName();
         try {
           final Object value = method.invoke(annotation, (Object[]) null);
-          parameters.put(method.getName(), value);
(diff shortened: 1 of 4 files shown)


codegen/src/main/java/io/sundr/codegen/functions/ClassTo.java
```diff
@@ -261,11 +261,12 @@ private static void processAnnotatedElement(AnnotatedElement field, List<Annotat
...
       Map<String, Object> parameters = new HashMap<>();
       for (java.lang.reflect.Method method : annotationType.getDeclaredMethods()) {
+        final String name = method.getName();
         try {
           final Object value = method.invoke(annotation, (Object[]) null);
-          parameters.put(method.getName(), value);
+          parameters.put(name, value);
         } catch (IllegalAccessException | InvocationTargetException e) {
-          throw new RuntimeException(e);
+          System.out.printf("Couldn't retrieve '%s' parameter value for %s%n", name, annotationType.getName());
```
iocanel, line 269: Can we address the `IllegalAccessException` in the first place?

metacosm: If you know how to do it, sure… :)

Issue #212, opened 2021-04-15
Revisit the implementation of equals(), hashCode() and toString() for basic types.


iocanel:
As of 0.30.0 some of the implementation of `equals`, `hashCode` and `toString` may have side effects that we need to eliminate.
Also we need to make sure that these implementations make sense. 

Some cases where we need to think about:

- Properties: should modifiers be part of equals/hashCode ?
- ClassRef: should generic arguments be part of equals/hashCode ?

There are definitely more, but these are some examples.

Issue #216, opened 2021-04-22
Code generation should not be bound to velocity.


iocanel:
Currently the code generation is bound to Apache velocity.
Over the years this has proven to limit our flexibility and control over the generated code and also is an additional transitive dependency we need to worry about.

Issue #213, opened 2021-04-15
Provide a stateless version of the model.


iocanel:
Currently, our domain object models use state, by making use of the `DefinitionRepository` under the hood. This means that calls on the exact same object may yield different results based on what's present on the repository.

While this is inevitable in some cases, I feel that at least of getters this should never be the case. In other words a getter of an immutable instance should return the same result.

Such cases are:

- `ClassRef.getDefinition()`
- `TypeDef.getOuterType()`

# Ideas

I would like our model to be usable with or without the `DefinitionRepository`. So, I'd like to have a jar/module that contains just the model without any reference to the `DefinitionRepository` or any other facility that maintains state. 

Why?

Because, having the state baked into the core model often leads to hard to troubleshoot issues like order sensitive operations etc.

On top of that jar/module we could provide Objects or tools that handle state.

## Options for model extensibility

Here are some ideas (not necessarily good ones)...

### Extending the types

We could have an additional jar/module that includes state-full implementations of the model. For example:

- `StatufullClassRef`
- `StatefullTypeDef`

While this approach is relatively easy from the user point of view, it still doesn't 100% address all issues (e.g. again we will have a single repo shared a static ref etc).

### Provide methods for mapping.
    
    public <T> T map(Function<ClassRef, T> mapper)

This could be used:

    ClassRef ref = ...
    TypeDef def = ref.map(toDefinition);

The benefit of this approach is that the repository instance to be used can be backed into the function we pass vs hard coded inside a `TypeDef` method. This makes it somehow easier to isolate the `state` (e.g. between tests).

The drawback is that we need take care of the mapping function (create it and pass it around).

PR #217, opened 2021-04-23
Make sundrio model stateless and idempotent

iocanel:
The goal of this pull request to make the sundrio model (`io.sundr.codegen.model`) stateless and idempotent.

### Background
Up until know model classes where using an repository that held all know type definitions (called `DefinitionRepository`).
Methods that we using this repo include:
- equals()
- hashCode()
- toString()

The undesirable side effect was that changing the content of the repository could alter the behavior of those classes, even when the internal state of the class remained the same. 

This is a code smell, that introduces a level of uncertainty in the sundrio API.

### Changes

- Introduce `render()` to replace `toString()` for code generation purposes.
- ClassRef and TypeDef model classes use no state:
    -  `ClassRef.getDefinition()` replaced by `GetDefinition.of()`
    - `TypeRef.isAssignable(TypeRef other)` replaced by `Assignable.isAssignable(TypeRef ref).from(TypeRef other)`
    - `TypeDef.isAssignable(TypeDef other)` replaced by `Assignable.isAssignable(TypeDef ref).from(TypeDef other)`
- ClassRef now uses an idempotent version of `getName()` and `getPackageName()`

### Related issues
Resolves: #212 #213
Related: #216 (It is not technically possible to proceed with 216, a followup pull request will deal with that).

### DIFF
--- a/annotations/builder/pom.xml
+++ b/annotations/builder/pom.xml
@@ -19,7 +19,7 @@
     <parent>
         <artifactId>annotations</artifactId>
         <groupId>io.sundr</groupId>
-        <version>0.30.5</version>
+        <version>0.30-SNAPSHOT</version>
     </parent>
     <modelVersion>4.0.0</modelVersion>
 
--- a/annotations/dsl/pom.xml
+++ b/annotations/dsl/pom.xml
@@ -19,7 +19,7 @@
     <parent>
         <artifactId>annotations</artifactId>
         <groupId>io.sundr</groupId>
-        <version>0.30.5</version>
+        <version>0.30-SNAPSHOT</version>
     </parent>
     <modelVersion>4.0.0</modelVersion>
 
--- a/annotations/pom.xml
+++ b/annotations/pom.xml
@@ -19,7 +19,7 @@
     <parent>
         <artifactId>sundr-pom</artifactId>
         <groupId>io.sundr</groupId>
-        <version>0.30.5</version>
+        <version>0.30-SNAPSHOT</version>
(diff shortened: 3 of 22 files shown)


geoand (approved):
I'm not the best person to judge this, but it sounds reasonable to me

metacosm:
Should we aim to use this change in the client's next release?

iocanel:
> Should we aim to use this change in the client's next release?

I feel that this change would add to stability of the pojo generator, since its now idempotency is guaranteed.

metacosm:
I tend to agree. What do you think @manusa, @rohanKanojia ?

manusa:
I like idempotent, stateless and reproducible. 
I don't know what the implications of this are for the client. i.e. Is this going to alter the generated code API in some way (disappearing methods)? or just the content of the methods? (Sorry for the maybe naive question)

iocanel:
> I like idempotent, stateless and reproducible.
> I don't know what the implications of this are for the client. i.e. Is this going to alter the generated code API in some way (disappearing methods)? or just the content of the methods? (Sorry for the maybe naive question)

The client generated code is not expected to change. The crd generator will have to align though due to the removal of `ClassRef.getDefinition()`. This is no biggie, I've aligned things locally and its trivial.

iocanel:
Not sure what exactly happened but the commits got lost after I rebased from the master

iocanel:
Created a new pull request that is a superset of this: https://github.com/sundrio/sundrio/pull/219

PR #218, opened 2021-04-23
feat: @ExternalBuildables can now include/exclude abstract classes an…

iocanel:
This pull request adds more control to the `@ExternalBuildables` annotation.

The intention is to use this annotation going forward in order to generate the model builders, without binding the model objects themselves with the `@Buildable` annotation. While experimenting with this option, I found  out that we did need this option.


geoand (approved):

aureamunoz (commented):
LGTM

PR #219, opened 2021-04-26
Module split

iocanel:
The goal of this pull request to make the sundrio model (io.sundr.codegen.model) stateless and idempotent.

Additionally, it extracts model to its own module and uses an `@ExternalBuildables` generator to generate fluent / builders. The generator is meant to be invoked manually and the generated files to be added to git (similar to how the kubernetes-client works).

Background
Up until know model classes where using an repository that held all know type definitions (called DefinitionRepository).
Methods that we using this repo include:

- equals()
- hashCode()
- toString()
The undesirable side effect was that changing the content of the repository could alter the behavior of those classes, even when the internal state of the class remained the same.

This is a code smell, that introduces a level of uncertainty in the sundrio API.

Also, we want to decouple the model from the adapters and make it something that one can use standalone.

Changes
- Introduce render() to replace toString() for code generation purposes.
- ClassRef and TypeDef model classes use no state:
- ClassRef.getDefinition() replaced by GetDefinition.of()
- TypeRef.isAssignable(TypeRef other) replaced by Assignable.isAssignable(TypeRef ref).from(TypeRef other)
- TypeDef.isAssignable(TypeDef other) replaced by Assignable.isAssignable(TypeDef ref).from(TypeDef other)
- ClassRef now uses an idempotent version of getName() and getPackageName()

Related issues
Resolves: #212 #213
Related: #216 (It is not technically possible to proceed with 216, a followup pull request will deal with that).


iocanel:
@metacosm: This is a clone of #217 with some added refactorings:

- make model a standalone module.
- model builders / fluent are now generated.

metacosm:
I don't know enough about the internals of sundrio to adequately review this PR. 

The thing that I'm perplex about, though, is that I'm not sure that we actually increase the idempotency of operations with this. As far as I can tell, the definition repository is still being used meaning that `GetDefinition.of` calls (for example) will still depend on the state of the repository at the time they're called. 

I'm not sure if it's feasible but what I'd really like to see is inclusion of the functionality that is currently found in the `Types` class of the fabric8 client as part of the contract of the `TypeDef`. More precisely, I'd rather see a stateful `TypeDef` (or another object that could fulfill this role) that could be "resolved/unshallowed" to have complete information, i.e. fully resolved typed (including generic ones) for the whole ancestor hierarchy. This operation would be blocking or failing until all the information is available and then be idempotent, so that a type would only need to go through that process once and store that operation result as part of itself so that it could be queried / passed around without needing to query the type system again.

Maybe this is the first step towards this goal, I don't know. I don't even know if that goal is part of sundrio's scope but that's what I'd like to see because this is what we need in the CRD generator (in particular) and, more generally, as part of what sundrio offers as, right now, the information we get is incomplete, at least from a client code point of view where the methods you can query all seem to be coming with asterisks…

The equivalent with Java types would be the difference between `get` vs `getDeclared` methods, I guess.

iocanel:
>As far as I can tell, the definition repository is still being used meaning that `GetDefinition.of` calls (for example) will still depend on the state of the repository at the time they're called.

That is correct, the `repo` hasn't been completely removed, it just doesn't affect core operations like `equals`, `hashCode` and `toString`. Having `repo` affect operations like these, violate the core guideliness, can potentially cause problems and adds to the uncertainty around the model.

Regarding `GetDefinition.of`, you are also correct... operations that need to invoke `GetDefinition.of` directly or indirectly do depend on the content of the `repo`. This is why we need to guarantee that internally we don't abuse this feature and only use it when it's absolutely needed. A huge part of this PR is about this.

From the CRD generator perspective, we need to ensure that the adapters (`ClassTo`, `ElementTo`) that populate the `repo` work correctly and go as deep as they need without falling into infinite loops. The PR doesn't directly contribute to that, but by making the model pretty much stateless and idempotent, makes it easier to narrow down issue and troubleshoot such issues.

> 
> More precisely, I'd rather see a stateful `TypeDef` (or another object that could fulfill this role) that could be "resolved/unshallowed" to have complete information, i.e. fully resolved typed (including generic ones) for the whole ancestor hierarchy. This operation would be blocking or failing until all the information is available and then be idempotent, so that a type would only need to go through that process once and store that operation result as part of itself so that it could be queried / passed around without needing to query the type system again.

That is something that we can possibly build on top of the current model.



> 
> Maybe this is the first step towards this goal, I don't know. I don't even know if that goal is part of sundrio's scope but that's what I'd like to see because this is what we need in the CRD generator (in particular) and, more generally, as part of what sundrio offers as, right now, the information we get is incomplete, at least from a client code point of view where the methods you can query all seem to be coming with asterisks…
> 
> The equivalent with Java types would be the difference between `get` vs `getDeclared` methods, I guess.

metacosm (approved):
As far as I can tell, this change moves in the right direction.

PR #222, opened 2021-04-28
refactor: Make assignable utils less confusing and prevent infinite loops

iocanel:


metacosm (approved):
The changes look reasonable to me…

PR #224, opened 2021-05-05
chore: only inline buildable types

iocanel:


metacosm (approved):
Looks reasonable to me.

PR #227, opened 2021-05-10
Split Adapters into multiple different modules

iocanel:
Resolves: #228 

The pull request does the following:

- Intdocues an adapter api (converters / function that convert java type representation into sundrio model).
- Introduce adapter impl
    - apt
    - reflection
    - sources (via github java parsers)

- Split utilities into:
   - core utilities
   - model utilities
   - adapter specific utilities

- Use the adapter api all accross the board.


adapters/api/src/main/java/io/sundr/adapter/api/Adapter.java
```diff
@@ -0,0 +1,52 @@
...
+import java.util.function.Function;
+
+import io.sundr.model.Method;
+import io.sundr.model.Property;
+import io.sundr.model.TypeDef;
+import io.sundr.model.TypeRef;
+
+public interface Adapter<T, R, P, M> {
+
+  Function<T, TypeDef> getFunction();
```
metacosm, line 29: Why not `getAdapterFunction` to be consistent with the other methods?

PR #239, opened 2021-05-20
Make velocity an optional codegen module

iocanel:
Resolves: #216 

The big picture around this pull request is to make Apache Velocity optional. Our model can be rendered into a String, without the need of an external rendering mechanism (e.g. a template engine). So our modules need to leverage the model strengths.

However, we have modules like the `transform` module and the `sundr-maven-plugin` that still need to use velocity. So, it needs to be an optional thing.

In detail the pull request:
- Introduces a new code generation API
- Provides APT related implementations to the code generation API
- Provide Velocity related implementations to the code generation API
- Align existing modules to use the new code generation API


iocanel:
@metacsom: Since the pull request does introduce a new general purpose API for code generation, I'd like to hear your opinion, as you recently implemented something similar for the `crd generator`.

codegen/api/src/main/java/io/sundr/codegen/api/CodeGenerator.java
```diff
@@ -0,0 +1,160 @@
...
+   * @return true if generation was succesful, false otherwise
+   */
+  public boolean generate(T... items) {
+    if (renderer == null) {
+      throw new IllegalStateException("Renderer should not be null.");
+    }
+
+    if (output == null) {
+      throw new IllegalStateException("Output should not be null.");
+    }
```
manusa, line 133: Not sure if this is stupid, but if these variables are final, why not throw the exception/perform validation in the constructor?

codegen/velocity/src/main/java/io/sundr/codegen/processor/TemplateGeneratingProcessor.java
```diff
@@ -106,8 +109,12 @@ public void generateFromStringTemplate(TypeDef model, String[] parameters, Strin
...
       return;
     }
-    generateFromStringTemplate(model, parameters, processingEnv.getFiler().createSourceFile(newModel.getFullyQualifiedName()),
-        content);
+
+    CodeGenerator.newGenerator(TypeDef.class)
+        .withIdentifier(TypeDef::getFullyQualifiedName)
+        .withOutput(new TypeDefAptOutput(processingEnv.getFiler()))
+        .withRenderer(VelocityRenderer.fromTemplate(content, parameters))
+        .generate(newModel);
```
manusa, line 117: Is the function in line 129 still used (`generateFromStringTemplate(T model, String[] parameters, FileObject fileObject, String content)`)?
If not, shouldn't it be marked for deprecation?

manusa (approved):
Besides the nit comments, and with my scarce knowledge of the project, everything LGTM

rohanKanojia (approved):

codegen/velocity/src/main/java/io/sundr/codegen/processor/TemplateGeneratingProcessor.java
iocanel: The class is removed.

Issue #248, opened 2021-06-09
Buildable fields in generic parents


shawkins:
This is in reference to https://github.com/fabric8io/kubernetes-client/issues/3076 

The CustomResource class holds generic references to the spec and status.  The CustomResourceFluentImpl generated by the project will have default handling for those fields because it's not known at that time if the spec/status classes will be Buildable.

If you then create a MyCustomResource extending CustomResource with spec/status classes that are buildable, there doesn't appear to be a way to override the default behavior - that is a Buildable MyCustomResourceFluentImpl will just defer to CustomResourceFluentImpl for the handling of spec/status.  That result is not desirable, because it is inconsistent with general fabric8 builder behavior that will deep clone the transitive buildable fields.

iocanel:
So the issue is that in this case we miss the nested fluent stuff, right?

That's a little tricky, but I'll have a look.

shawkins:
Right, it will copy by reference rather than via a builder.  That also means you miss out on a lot of fluent methods, you instead have to do ```new MyCustomResourceBuilder().withSpec(new MyCustomResourceSpecBuilder()...```

iocanel:
Unfortunately, the `Generated` annotaiton was introduced in jdk 9.
So, for now we will just add a comment instead.

iocanel:
I will have a look at it ASAP

iocanel:
@shawkins: I created #260 which should provide the desired user experience.

Can you have a look and provide some feedback before I merge/release?

shawkins:
@iocanel thank you.  Yes, I'll pull this locally and try it out.

shawkins:
The generated code looks good.  ConnectionFluentImpl assumes the responsibility of the nested spec.  The only difference with the fabric8 structure is that the base class will be in a different module/package - but I don't think that should matter.

iocanel:
> The generated code looks good. ConnectionFluentImpl assumes the responsibility of the nested spec. The only difference with the fabric8 structure is that the base class will be in a different module/package - but I don't think that should matter.

I don't think that this will be an issue.

PR #251, opened 2021-06-22
Introdcue RichTypeDef and utilities for applying type parameters

iocanel:
The idea of this pull request is to introduce `RichTypeDef` that provides access to super class methods.
In addition, it does provide utilities for applying type parameters to a `TypeDef`.

This is the first step towards #248


iocanel:
@metacosm: In the past we discussed about something similar, so It would be nice to have your opinion.

metacosm (commented):
I like having the `RichTypeDef` option and the rest of the code looks OK to me at a cursory glance.

PR #269, opened 2021-07-28
Misc fixes related to apt

iocanel:
This pull request aims at addressing the following issues:

- Rendering issues related to annotations
- Providing reasonable error messages when apt error related to unresolved symbols occur

### DIFF
--- a/annotations/builder/src/main/java/io/sundr/builder/internal/utils/BuilderUtils.java
+++ b/annotations/builder/src/main/java/io/sundr/builder/internal/utils/BuilderUtils.java
@@ -356,7 +356,10 @@ public class BuilderUtils {
       try {
         result.add(context.getElements().getTypeElement(ref.value().getCanonicalName()));
       } catch (MirroredTypeException e) {
-        result.add(context.getElements().getTypeElement(e.getTypeMirror().toString()));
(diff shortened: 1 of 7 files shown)


annotations/builder/src/main/java/io/sundr/builder/internal/utils/BuilderUtils.java
```diff
@@ -356,7 +356,10 @@ public static TypeDef getInlineReturnType(BuilderContext context, Inline inline,
       try {
         result.add(context.getElements().getTypeElement(ref.value().getCanonicalName()));
       } catch (MirroredTypeException e) {
-        result.add(context.getElements().getTypeElement(e.getTypeMirror().toString()));
+        Element element = context.getTypes().asElement(e.getTypeMirror());
```
metacosm, line 359: Shouldn't we extract the body of this method so that it's shared between both methods?

metacosm (approved):
LGTM apart from the possible refactoring opportunity.

annotations/builder/src/main/java/io/sundr/builder/internal/utils/BuilderUtils.java
iocanel: Yeah, I think that at some point we should.

Issue #279, opened 2021-09-22
Code generator looks for existing classes using current classloader


laurentgo:
The base class for code generators uses a `classExists` method in order to confirm if the class needs to be generated or not.
The method uses the current classloader to perform the check, but this assumes that the javac compiler either relies on the default discovery mechanism for annotation processors (meaning searching in all source dependencies for annotation processors), or that the annotation processor classpath contains also the dependencies.

Instead the code generator should query the compiler using the `Elements` helper to look for the class using the `getTypeElement()` method.

iocanel:
The apt specific implementation looks like: https://github.com/sundrio/sundrio/blob/master/adapters/apt/src/main/java/io/sundr/adapter/apt/AptLookup.java

Shouldn't that be enough?

laurentgo:
Sorry, I was referring to this class: https://github.com/sundrio/sundrio/blob/7d2c100a8cd3fdaa0b326c006f895a06066c5134/annotations/builder/src/main/java/io/sundr/builder/internal/processor/AbstractBuilderProcessor.java#L89

iocanel:
> Sorry, I was referring to this class:
> 
> https://github.com/sundrio/sundrio/blob/7d2c100a8cd3fdaa0b326c006f895a06066c5134/annotations/builder/src/main/java/io/sundr/builder/internal/processor/AbstractBuilderProcessor.java#L89

You are right! I think that this needs to change and make use of `Elements` as you suggested.
I'll fix that asap, unless you prefer to create a PR yourself (no pressure) :smile:

laurentgo:
PR submitted

PR #280, opened 2021-09-22
Use Elements helper to check if a class already exists

laurentgo:
To decide if a class has to be generated or has to be skipped because it
already exists in the classpath, the processor uses
`AbstractBuilderProcessor#classExists(String)` to check for the class
existence.

Currently the method relies on the class being present in the classpath,
but that presupposes that the annotation processor classpath is the same
as the source classpath, which is not always the case.

Changes the method to rely instead on the Elements class helper to
inspect the source classpath.

Fixes sundrio/sundrio#279


iocanel (approved):
LGTM

PR #282, opened 2021-09-24
fix: properly detect names if '{' is not on the same line

### LINKED ISSUE #281: `Types.parseName` doesn't work if the '{' is not on the same line as the declaration

metacosm:
Fixes #281


iocanel (approved):
LGTM

PR #286, opened 2021-09-30
fix: ensure that only actual java source end up under `generated-sour…

iocanel:
…ces`.


iocanel:
Since #287 was merged, I think that this needed to close.

PR #296, opened 2022-01-06
Fix Index out-of-bound in Collection/Map/Optional argument resolution

andreaTP:
Reference: https://github.com/fabric8io/kubernetes-client/issues/3716

### DIFF
--- a/annotations/builder/src/main/java/io/sundr/builder/internal/functions/TypeAs.java
+++ b/annotations/builder/src/main/java/io/sundr/builder/internal/functions/TypeAs.java
@@ -21,7 +21,9 @@ import static io.sundr.model.utils.Types.*;
 
 import java.util.ArrayList;
 import java.util.List;
+import java.util.Optional;
 import java.util.function.Function;
+import java.util.stream.Stream;
 
 import javax.lang.model.element.Modifier;
 
@@ -29,6 +31,7 @@ import io.sundr.FunctionFactory;
 import io.sundr.builder.internal.BuilderContext;
 import io.sundr.builder.internal.BuilderContextManager;
 import io.sundr.model.*;
+import io.sundr.model.functions.GetDefinition;
 import io.sundr.model.utils.Collections;
 import io.sundr.model.utils.Types;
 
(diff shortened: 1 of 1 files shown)


annotations/builder/src/main/java/io/sundr/builder/internal/functions/TypeAs.java
```diff
@@ -216,20 +219,32 @@ public ClassRef apply(TypeRef item) {
...
 
+  private static Optional<TypeRef> extractArgument(ClassRef classRef, Function<TypeRef, Boolean> typeCheckFn,
+      int argumentIndex) {
+    if (typeCheckFn.apply(classRef)) {
+      if (classRef.getArguments().size() > argumentIndex) {
+        return Optional.of(classRef.getArguments().get(argumentIndex));
+      } else {
+        // Fallback
+        TypeDef typeDef = GetDefinition.of(classRef);
+        return Stream
```
iocanel, line 235: Since we go down that road, then maybe we should check the whole hierarchy instead of classes directly extended / implemented. 

It's possible that getting the `RichTypeDef` can easily address that.

iocanel (changes requested):
LGTM, just requrested a small change so that we cover all cases.

annotations/builder/src/main/java/io/sundr/builder/internal/functions/TypeAs.java
andreaTP: Thanks a lot for the feedback @iocanel , I have attempted to follow the suggestion but I might be missing some context (sorry I'm not really familiar with the codebase).

Let me try to understand:
 - what invariants are we missing by recursively iterating on extended/implemented classes? My understanding is that we are covering the full hierarchy
 - `RichTypeDef` doesn't seems to directly offer methods/helpers to traverse parents, do you have a reference or example of what you are thinking about?
 - The closest thing I found in the codebase is `Types.visitParents` but I couldn't get it to work as expected

andreaTP:
A more general question @iocanel , once we agree on the implementation details will it be possible to backport the fix and release a patch of the `0.50` series?
Thanks a lot in advance!

iocanel:
> A more general question @iocanel , once we agree on the implementation details will it be possible to backport the fix and release a patch of the `0.50` series? Thanks a lot in advance!

If there is justification for a 0.50.x we'll cut one.
The question is why not bump the client to the latest release?

andreaTP:
That would be ideal, unfortunately we found out that 0.80 contains breaking changes ( ref https://github.com/fabric8io/kubernetes-client/pull/3718 ) and is unlikely to be merged before 5.12 of kubernetes-client get released.

I'm trying hard to have a bunch of fixes in time for 5.12 to be able to use the new CRD generator for the new Keycloak operator. 🙂

iocanel:
> That would be ideal, unfortunately we found out that 0.80 contains breaking changes ( ref [fabric8io/kubernetes-client#3718](https://github.com/fabric8io/kubernetes-client/pull/3718) ) and is unlikely to be merged before 5.12 of kubernetes-client get released.
> 
> I'm trying hard to have a bunch of fixes in time for 5.12 to be able to use the new CRD generator for the new Keycloak operator. slightly_smiling_face

Created 0.50.x branch, I'll merge and cherry-pick to 0.50.x

andreaTP:
@iocanel that's brilliant! Thanks a ton! Let me know when is out 🙂 ❤️

iocanel:
> @iocanel that's brilliant! Thanks a ton! Let me know when is out slightly_smiling_face heart

0.50.3 is already on central :wink:

andreaTP:
Wow I owe you a beer 🍻

Issue #307, opened 2022-03-11
Buildable annotation should take into account parent(abstract) class fields


andreaTP:
The `Buildable` annotation should take into account and use fields defined in the ancestor classes.

Here you have a reproducer:
https://github.com/andreaTP/sundrio_abstract_builder

clone the repo and run:
`mvn clean compile`

We expect that the `Main` class compiles leveraging the generated classes, but it doesn't, since the fields defined in the `abstract` ancestor are not taken into account.

andreaTP:
cc. @iocanel

PR #311, opened 2022-03-17
Fix handling of inherited fields from non buildable parents

iocanel:
Resolves: #307


iocanel:
@andreaTP: I think that it was failing to recognize the setter of the generic property. It should work fine now.

andreaTP:
@iocanel tested, thanks for fixing this!

PR #315, opened 2022-03-30
feat: make it possible to express Visitor requirements

iocanel:


iocanel:
The pull request introduces a new method on the visitor interface:

```
  public Predicate[] getRequirements();
```

A visitor will only visit `if and only if` all predicates are satisfied by at least one node in the visitor path.

PR #326, opened 2022-06-07
chore: Change distribution from adopt to temurin

jorsol:
Adopt OpenJDK got moved to Eclipse Temurin and won't be updated anymore.


iocanel (approved):
LGTM

PR #330, opened 2022-07-05
feat: use Entry<String, Object> to capture path.

iocanel:
The pull request changes the type of path from list of objects to a list of entries String -> Object. This makes it possible to also track the exact route of visited nodes prior to the current.


iocanel:
@manusa: I'd like that to be part of 6.0 if possible.

manusa:
I think that this shouldn't be breaking, so there should be no problem including the version bump.
This one wasn't merged because it didn't compile
https://github.com/fabric8io/kubernetes-client/pull/4212 (there seems to be an issue with the CRD Generator test expectations)

iocanel:
@manusa: pushed a fix that will take care of these issues. Releasing soon.

manusa:
> @manusa: pushed a fix that will take care of these issues. Releasing soon.

Great thx!

PR #334, opened 2022-09-07
Fix issues with cyclic type argumetns

iocanel:


iocanel:
@xRodney: Thank you so much for the analysis and the test.
@andreaTP: This should address the issues you had.

andreaTP:
@iocanel thanks a ton!!! Can you please notify when a new release will be available to test?

Issue #333, opened 2022-09-06
StackOverflowError on TypeArguments.apply(typeDef) for MultiMaps & other non-trivial generics


xRodney:
Hello,

I ran into this error while working on https://github.com/fabric8io/kubernetes-client/issues/4357. As written there, sundrio is looping infinitely, on some more complicated generic classes.

An example of such class is a multi map:
```
public class MultivaluedHashMap<K, V> extends HashMap<K, List<V>> { /* ...  */ }
```

In kubernetes-client, the code involved in the loop is in part custom to that project, which is why at first I thought the error is there. However, then I found almost identical code is also present in sudrio project. 
In both cases, the code is responsible for "inflating" TypeDeft to include also members from their superclasses. 
In sundrio, this logic is implemented in an util `TypeArguments`
So I wondered if it suffers from the same bug and sure enough, it stack overflows as well.

Here's a minimal failing example:
```
package io.sundr.model.utils;

import org.junit.Test;

import io.sundr.model.ClassRef;
import io.sundr.model.Kind;
import io.sundr.model.TypeDef;
import io.sundr.model.TypeDefBuilder;
import io.sundr.model.functions.GetDefinition;
import io.sundr.model.repo.DefinitionRepository;

import static io.sundr.model.utils.Collections.K;
import static io.sundr.model.utils.Collections.V;

public class MultiMapTest {
  public static final TypeDef MULTI_HASH_MAP = new TypeDefBuilder(TypeDef.forName("com.example.MultiHashMap"))
          .withKind(Kind.CLASS)
          .withParameters(K, V)
          .withExtendsList(Collections.HASH_MAP.toReference(K.toReference(), Collections.LIST.toReference(V.toReference())))
          .build();

  @Test
  public void testName() {
    DefinitionRepository.getRepository().register(Collections.MAP);
    DefinitionRepository.getRepository().register(Collections.HASH_MAP);
    DefinitionRepository.getRepository().register(Collections.LIST);
    DefinitionRepository.getRepository().register(Types.STRING);
    DefinitionRepository.getRepository().register(Types.INT);
    DefinitionRepository.getRepository().register(MULTI_HASH_MAP);

    ClassRef classRef = MULTI_HASH_MAP.toReference(Types.STRING_REF, Types.INT_REF);
    TypeDef typeDef = GetDefinition.of(classRef);

    System.out.println(classRef.render());
    System.out.println(typeDef.render());

[...]

PR #340, opened 2022-09-12
TypeArguments: eliminate StackOverflow while retaining accurate types

xRodney:
Fixes sundrio/sundrio#333.

While the original fix replaced the recursive mapping `V -> List<V>` with `V -> List<?>`, here I take a more subtle approach. The mapping is replaced with `V -> List<V>[replaced=true]`, where he `[replaced=true]` is an attribute. The visitor never replaces class refs with this attribute and while doing so also removes it, so that the resulting type tree is clean.

Because the goal of this is to be usable in `crd-generator`, I also added a new convenience method `RichTypeDef apply(ClassRef)` (I actually renamed a pre-existing private method with the same name to make room for it). This can process the types all the way from a generic reference (as opposed to from a definition)

Related: fabric8io/kubernetes-client#4357

### DIFF
--- a/model/utils/src/main/java/io/sundr/model/utils/TypeArguments.java
+++ b/model/utils/src/main/java/io/sundr/model/utils/TypeArguments.java
@@ -99,6 +99,10 @@ public class TypeArguments {
     }
 
     List<TypeParamDef> parameters = definition.getParameters();
+    if (parameters.size() != arguments.size()) {
+      throw new IllegalStateException("Incompatible reference " + ref + " to " + definition);
+    }
+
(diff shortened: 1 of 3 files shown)


xRodney:
Hello @iocanel and thanks a lot for your review. Would you please mind merging and releasing it?

xRodney:
Hello again @iocanel,
I realize you might be busy, but is there a chance you could have another look on this? Or at least a rough estimate when you'd be less busy? We are blocked by kubernetes-client#4357.
Thanks a lot.

PR #344, opened 2022-10-08
Fix build: Examples should have the 0.93-SNAPSHOT

xRodney:
same as the project, they were one version late.


iocanel:
This has been already fixed.

PR #358, opened 2023-02-15
Add maven.clean.skip=true to generate BOM

radcortez:
The property used to skip the clean was renamed to `maven.clean.skip` in the clean plugin 3.x. This just set both properties to support each version.

Check: https://maven.apache.org/plugins/maven-clean-plugin/clean-mojo.html#skip


iocanel (approved):
LGTM

PR #367, opened 2023-05-08
fix #357: allowing for null instance with builders

### LINKED ISSUE #357: Add a convenience method for constructing a builder from a null value
In code paths that do some kind of create or update logic it's common to have an optional existing value.  Constructing the builder currently with a null instance results in an exception.  Can either null be allowed or a static from(instance) method be introduced?

shawkins:
Addresses #357 by allowing null instance values in the builder constructors.

### DIFF
--- a/annotations/builder/src/main/java/io/sundr/builder/internal/functions/ToMethod.java
+++ b/annotations/builder/src/main/java/io/sundr/builder/internal/functions/ToMethod.java
@@ -318,6 +318,9 @@ class ToMethod {
       }
 
       if (!descendants.isEmpty()) {
+        statements.add(new StringStatement(
+            "if (" + argumentName + "==null){ this." + fieldName + " = null; _visitables.remove(\"" + fieldName + "\");"
+                + " return (" + returnType + ") this;}"));
         for (Property descendant : descendants) {
           ClassRef dunwraped = new ClassRefBuilder(
               (ClassRef) combine(UNWRAP_COLLECTION_OF, UNWRAP_ARRAY_OF, UNWRAP_OPTIONAL_OF).apply(descendant.getTypeRef()))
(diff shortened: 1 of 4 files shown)


annotations/builder/src/main/java/io/sundr/builder/internal/functions/ToMethod.java
```diff
@@ -318,18 +318,25 @@ private List<Statement> getStatements(Property property, List<ClassRef> alsoImpo
...
           statements.add(new StringStatement("if (" + argumentName + " instanceof " + dunwraped + "){ this." + fieldName
               + "= new " + builderRef + "((" + dunwraped + ")" + argumentName + "); _visitables.get(\"" + fieldName
-              + "\").add(this." + fieldName + ");}"));
-
+              + "\").clear(); _visitables.get(\"" + fieldName + "\").add(this." + fieldName + ");}"));
           alsoImport.add(dunwraped);
           alsoImport.add(builderRef);
         }
+        statements.add(new StringStatement(
+            "else {  VisitableBuilder<? extends " + ((ClassRef) type).getFullyQualifiedName()
```
iocanel, line 336: I think that we can eliminate the `else` here as the if returns.

iocanel (approved):
LGTM. 
There is an optional comment that could be addressed if it makes sense.

iocanel:
@shawkins: This will also need a rebase and conflict resolution.

annotations/builder/src/main/java/io/sundr/builder/internal/functions/ToMethod.java
shawkins: Removed now.

shawkins:
> This will also need a rebase and conflict resolution.

Should be good now.

PR #368, opened 2023-05-09
fix #263 merging fluent / fluentimpl

### LINKED ISSUE #263: Eliminate Fluent interface
Historically, we separated the `Builder` from the `Fluent`/`FluentImpl` in order to supportthe `Inline` concept: `Being able to use the fluent part of a builder, but with having a different method in the end instead of build. An example of how it was used in the `kubernetes-client`:

```java
Pod pod = client.pods().createNew()
   .withNewSpec()
      .withName("my-pod")
  .endSpec()
  // more code
  .done();
```

It seems though that we don't need both a `Fluent` interface and a `FluentImpl` class.
It might make sense if we collapsed them into one class that will be called `Fluent`.

`Builder` heavy libraries and frameworks could be greatly benefited by the reduced size and the simplified signatures.

shawkins:
Hopefully this is still valid.

There is a usage of the fluent interface/impl in fabric8 that can be easily changed to just the concrete fluent.

cc @iocanel @manusa

### DIFF
--- a/annotations/builder/src/main/java/io/sundr/builder/internal/functions/ClazzAs.java
+++ b/annotations/builder/src/main/java/io/sundr/builder/internal/functions/ClazzAs.java
@@ -47,7 +47,6 @@ import io.sundr.model.AnnotationRef;
 import io.sundr.model.AnnotationRefBuilder;
 import io.sundr.model.ClassRef;
 import io.sundr.model.ClassRefBuilder;
-import io.sundr.model.Kind;
 import io.sundr.model.Method;
 import io.sundr.model.MethodBuilder;
 import io.sundr.model.Modifiers;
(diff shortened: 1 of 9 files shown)


iocanel:
@shawkins: Thank you for putting time on this. This is a far from trivial task. In the past I tried myself and gave up. 

If I remember correctly the major pain point for me was what to do with intrefaces annotated with `@Buildable` as ATM we do generate a `Fluent` that is implemented by Builders of classes implementing the original interface.

Eitherway, before we merge this. I need to ensure that all downstream projects work. So, it will take some time.

shawkins:
@iocanel would it be possible to include additional test cases for the things that may be problematic?

I tried with a buildable interface and that caused the pojo to be generated as well.  From there the fluent and builder looked as expected.

With an abstract buildable I see only the fluent being generated.  Just like the change needed in fabric8, downstream projects would no longer use the fluent interface and just extend the concrete fluent.

iocanel:
> @iocanel would it be possible to include additional test cases for the things that may be problematic?

I think that the `shapes` example should cover most of cases.

> I tried with a buildable interface and that caused the pojo to be generated as well. From there the fluent and builder looked as expected.

Before adding support for `Pojo` generation when interfaces where annotated a `Fluent` was generated.
Implementing types had their Fluent extend the interface Fluent (or so I think). This seems to be broken now anyways and most probably this isn't even used.

I do see some failures in the downstream check which are due to explict references on the Impl class.
Can you please address that downstream so that we can re-run the check.

shawkins:
> Can you please address that downstream so that we can re-run the check.

Won't I need at least a snapshot release of sundrio before I can do that?

shawkins:
@iocanel how do you want to proceed on this?

iocanel:
> > Can you please address that downstream so that we can re-run the check.
> 
> Won't I need at least a snapshot release of sundrio before I can do that?

I think that you might be able to fix that without a new release just by changing from FluentImpl to Fluent.

shawkins:
> I think that you might be able to fix that without a new release just by changing from FluentImpl to Fluent.

How would that work given that Fluent with existing releases of sundrio will be an interface?  

I'll just wait for the new release and update then.

iocanel:
> How would that work given that Fluent with existing releases of sundrio will be an interface?
> 

Well, unless you explicitly implment the `Fluent` I don't see why that is a problem.

shawkins:
State before this pr:

fabric8 has RunConfigNestedImpl extends RunConfigFluentImpl<RunConfigNested> implements RunConfigNested - https://github.com/fabric8io/kubernetes-client/blob/5fb4ea82dfe55932a10b646fecde7812b0a28ef0/kubernetes-client-api/src/main/java/io/fabric8/kubernetes/client/extended/run/RunOperations.java#L31

After this pr, RunConfigFluentImpl will no longer exist.  The fabric8 code will to instead become:

RunConfigNested extends RunConfigFluent<RunConfigNested>

I cannot make that change ahead of time as RunConfigFluent is an interface.

iocanel:
Ok, I am creating a 0.95.0 release with all the other changes that have recently made it in.
Then I'll branch and we can go for 0.100.0.

shawkins:
> Then I'll branch and we can go for 0.100.0.

Sounds good.  Will there be snapshot builds available to update the downstream against?

There are some additional changes that may be good before the final 0.100.0 release.  Please see https://github.com/sundrio/sundrio/pull/379/commits/6168288168615b5bdf691dc4b25e842aa878e6dc

iocanel:
> Sounds good. Will there be snapshot builds available to update the downstream against?

I haven't configured snapshot releases. Will work on that.

iocanel (approved):
LGTM

Issue #350, opened 2022-11-18
Builder methods with interface types silently ignore some concrete types


shawkins:
Relates to https://github.com/fabric8io/kubernetes-client/issues/4540 - if the concrete type is not specifically registered as buildable, then calling the Builder.withInterface(concreteValue) method will just silently ignore the value.  At the very least an exception should be thrown.

PR #375, opened 2023-05-11
fix #374 updating method logic to address several issues

### LINKED ISSUE #374: Various fluent / builder issues
There are several more fluent issues:

1. removeAll usage is not correct (wrong type), it should be clear instead:
```
public A withParameters(List<Parameter> parameters) {
    if (this.parameters != null) { _visitables.get("parameters").removeAll(this.parameters);}
    ...
}
```

2. setTo and addTo with an index and an parent / interface type do not have fallbacks, e.g. 

setToCollection(Integer index,InterfaceType item)

3. with varargs does not clear the visitables
```
  public A withParameters(io.fabric8.openshift.api.model.Parameter... parameters) {
    if (this.parameters != null) {this.parameters.clear();}
    if (parameters != null) {for (Parameter item :parameters){ this.addToParameters(item);}} return (A) this;
  }
```

4. addTo with index and visitablebuildable does not use the same indexing rules as the other methods that allow for out-of-bounds values:

```
  public A addToObjects(Integer index,VisitableBuilder<? extends HasMetadata,?> builder) {
    if (this.objects == null) {this.objects = new ArrayList<VisitableBuilder<? extends HasMetadata,?>>();}
    _visitables.get("objects").add(index, builder);this.objects.add(index, builder); return (A)this;
  }
```

shawkins:


annotations/builder/src/main/java/io/sundr/builder/internal/functions/ToMethod.java
```diff
@@ -813,6 +817,14 @@ private Statement createAddToDescendantsFallback(String type, String name) {
               "else {  VisitableBuilder<? extends " + type + ",?> builder = builderOf(item); _visitables.get(\"" + name
                   + "\").add(builder);this." + name + ".add(builder); }");
         }
+
+        private StringStatement createAddOrSetIndex(String op, String propertyName, String returnType) {
+          return new StringStatement(
+              "if (index < 0 || index >= " + propertyName + ".size()) { _visitables.get(\"" + propertyName
```
iocanel, line 823: Can you please elaborate here?
Why do we need fallbacks? Shouldn't we just let it fail instead?

shawkins: > Why do we need fallbacks? Shouldn't we just let it fail instead?

The fallbacks, via builderOf, handles the failure case.  Several these methods had the same problem as #350

iocanel: I am still not sure I follow. Can you please further elaborate?
Given that we also have an additional topic regarding Editables in #376 it might make sense to schedule a meeting or something.

shawkins: > I am still not sure I follow. Can you please further elaborate?

The existing implementation of setToCollection(index, interface) looks like:

```
if (item instanceof Concrete1) { setToConcrete...
if (item instanceof Concrete2) { setToConcrete...

return (A)this;
```

So if the item is a not a known buildable type it will simply fall through and do nothing - which is what was observed with #350

This change makes setTo and addTo have the else case

} else { VisitableBuilder builder = builderOf(item) ... }

So that if something has it builder it will succeed, or if not there will be an exception.

annotations/builder/src/main/java/io/sundr/builder/internal/functions/ToMethod.java
```diff
@@ -744,12 +740,20 @@ public List<Method> apply(final Property property) {
 
             addSingleItemAtIndex = new MethodBuilder(addSingleItemAtIndex).withParameters(parameters).editBlock()
                 .withStatements(createAddToDescendants("addTo", descendants, true),
-                    new StringStatement("return (" + returnType + ")this;"))
```
iocanel, line 747: There is a method available caleld `createAddToDescendantsFallback` maybe this should be used instead?
It seems as `createAddToDescendantsFallback` was meant to do the same thing but somehow got forgotten.

shawkins: > There is a method available caleld createAddToDescendantsFallback maybe this should be used instead?

To clarify there are two things happening.  One is usage of builderOf, the other is the actual operation.  createAddToDescendantsFallback is written to do both, so it's not applicable to the other operations - indexed addTo, setTo, remove.

Note that not everywhere createAddOrSetIndex is a fallback case - the same logic applies to both fallback https://github.com/sundrio/sundrio/pull/375/files#diff-f95a091f7367292c3d8e6c10eb03b6411bfe67573270d10bb9393f947dc5ec54R755 and regular logic https://github.com/sundrio/sundrio/pull/375/files#diff-f95a091f7367292c3d8e6c10eb03b6411bfe67573270d10bb9393f947dc5ec54R773 - this does ensure that consistent indexing is used across the methods, which is also called out in #374

iocanel (approved):
lgtm

PR #376, opened 2023-05-16
minor changes related to Editable

shawkins:
correcting the readme and looking at both the context and the class's classloader for the builder.

@iocanel @manusa on the Editable interface, currently there's need for it to be generic, it could just as easily be defined as:

```
public interface Editable {

  VisitableBuilder<?, ?> edit();
}
```
or to match the behavior in BaseFluent.builderOf
```
public interface Editable<T> {

  VisitableBuilder<T, ?> edit();
}
```
However either of these would likely be considered breaking changes, so I'm not sure if it's worth pursuing.

### DIFF
--- a/annotations/builder/readme.md
+++ b/annotations/builder/readme.md
@@ -330,7 +330,7 @@ Circle circle = new CircleBuilder()
                       .build();
 
 // Create a new CircleBuilder with edit(), set props and the build
-Circle clone = ((Editable) circle).edit().withRadius(120).build();
+Circle clone = ((EditableCircle) circle).edit().withRadius(120).build();
 ```
 
 You can disable this behaviour by using `@Buildable(editableEnabled =
--- a/core/src/main/java/io/sundr/builder/BaseFluent.java
+++ b/core/src/main/java/io/sundr/builder/BaseFluent.java
@@ -40,10 +40,15 @@ public class BaseFluent<F extends Fluent<F>> implements Fluent<F>, Visitable<F> 
     }
 
     try {
-      return (VisitableBuilder<T, ?>) Class.forName(item.getClass().getName() + "Builder").getConstructor(item.getClass())
+      return (VisitableBuilder<T, ?>) Class.forName(item.getClass().getName() + "Builder", true, item.getClass().getClassLoader()).getConstructor(item.getClass())
           .newInstance(item);
     } catch (Exception e) {
-      throw new IllegalStateException("Failed to create builder for: " + item.getClass(), e);
+      try {
+        return (VisitableBuilder<T, ?>) Class.forName(item.getClass().getName() + "Builder").getConstructor(item.getClass())
+            .newInstance(item);
+      } catch (Exception e1) {
+        throw new IllegalStateException("Failed to create builder for: " + item.getClass(), e1);
+      }
     }
   }
 
@@ -63,6 +68,7 @@ public class BaseFluent<F extends Fluent<F>> implements Fluent<F>, Visitable<F> 
     return new LinkedHashSet(Arrays.stream(sets).filter(Objects::nonNull).collect(Collectors.toSet()));
   }
 
+  @Override
   public F accept(Visitor... visitors) {
     return accept(Collections.emptyList(), visitors);
   }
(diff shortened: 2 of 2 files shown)


shawkins:
> The suggested change restricts the return type of the edit method to VisitableBuilder without providing the ability specify a concrete builder, which defeats the purpose.

With either of the proposed signatures, the concrete class would still be:

```
  public ConcreteBuilder edit() {
    return new ConcreteBuilder(this);
  }
```

Maybe I'm missing something, but when I'm already dealing with the concrete class and the edit method signature is ConcreteBuilder edit(), there's really no need for having the generic parameterized by the builder.

> This way the only thing one could do with edit is to call build

That is correct, which is exactly the usage pattern for cloning and builderOf.  The current logic, because of a lack of bounds, doesn't even make an assumption about what the edit method returns:

```
    if (item instanceof Editable) {
      Object editor = ((Editable) item).edit();
      if (editor instanceof VisitableBuilder) {
        return (VisitableBuilder<T, ?>) editor;
      }
    }
```
with "VisitableBuilder<T, ?> edit()", this instead becomes
```
    if (item instanceof Editable) {
      return ((Editable<T>) item).edit();
    }
```
or a user / fabric8 looking to clone would do
```
    if (item instanceof Editable) {
      return ((Editable<T>) item).edit().build();
    }
```

iocanel:
> With either of the proposed signatures, the concrete class would still be:
>
> ```
>   public ConcreteBuilder edit() {
>     return new ConcreteBuilder(this);
>   }
> ```

Right! The implementing methods return type **can** be a subtype.

> Maybe I'm missing something, but when I'm already dealing with the concrete class and the edit method signature is ConcreteBuilder edit(), there's really no need for having the generic parameterized by the builder.
> 
> > This way the only thing one could do with edit is to call build
> 
> That is correct, which is exactly the usage pattern for cloning and builderOf.

Why not just go for a `clone`/`copy` method instead? Calling `edit` without editing seems a little weird to me.

> The current logic, because of a lack of bounds, doesn't even make an assumption about what the edit method returns:
> 
> ```
>     if (item instanceof Editable) {
>       Object editor = ((Editable) item).edit();
>       if (editor instanceof VisitableBuilder) {
>         return (VisitableBuilder<T, ?>) editor;
>       }
>     }
> ```
> 
> with "VisitableBuilder<T, ?> edit()", this instead becomes
> 
> ```
>     if (item instanceof Editable) {
>       return ((Editable<T>) item).edit();
>     }
> ```

I am missing the context here.

shawkins:
> Right! The implementing methods return type can be a subtype.

But in practice they won't be - as far as I can understand classes that implement Editable will be generated so the concrete signature will always be as expected.

> I am missing the context here.

This is being looked at under https://github.com/fabric8io/kubernetes-client/issues/5135 and the prospective changes to our generation logic https://github.com/fabric8io/kubernetes-client/pull/5138 - but I understand that changing the generic parameters of Editable or the signature of edit may not be worth considering.

In the current BaseFluent logic - https://github.com/sundrio/sundrio/blob/259219a10376a1beaaf5191306cd22c10b5706ea/core/src/main/java/io/sundr/builder/BaseFluent.java#L35 - is doing an instanceOf and an unchecked cast.  I was trying to at least avoid the instanceOf and/or simply the Editable interface.  You could of course just do something like 

```
public interface Editable<T extends VisitableBuilder<?, T>> {

  T edit();
}
```
to stick with the current interface and provide a little additional typing.

> Why not just go for a clone/copy method instead? Calling edit without editing seems a little weird to me.

That's fine with me as well, I presume that would added for now as a default method on Editable correct?  Unless the generic parameters were expanded to include the concrete type, copy would have to simply return Object.

Alternatively edit may not right method name - with lombok this method would normally be called toBuilder so it doesn't necessarily connote editing.

iocanel:
I think that we could possibly avoid the unchecked cast by moving the `builderOf` method from `BaseFluent` to concrete Fluent. This way we would possibly avoid casting, reflections etc.

shawkins:
> I think that we could possibly avoid the unchecked cast by moving the builderOf method from BaseFluent to concrete Fluent. This way we would possibly avoid casting, reflections etc.

Let's table any changes to the Editable interface and the builderOf method to a separate issue / pr, so that we can at least these base changes in.

iocanel (approved):
lgtm

PR #379, opened 2023-05-26
Post fluent / fluentImpl changes

shawkins:
This won't be ready for main until after #263 has been merged - the second commit are the new changes.

@iocanel I see that the changes to have fallbacks everywhere are incomplete - several locations do not do proper null handling of the collection field.  However I would like to propose rather than simply adding/refactoring those in, to simplify the methods to consolidate where the builder is coming from.

To assist with that these changes include a lookup table for all of the known types to avoid reflection for known types, but from what I can see locally you typically save less than 1 microsecond per invocation when the type is found in the instanceof chain or this lookup table vs calling builderOf.  So it would be fine to not even use this additional lookup method as well.

With that there are also several deprecations to the subtype methods.  For all of these methods there is a more general method that will work.  In particular removeMatchingFromXXXCollection methods should just use the removeMatchingFromCollection method - there's no conversion to a bulider involved there at all.  With the other consolidation of the builder logic, I would go further and deprecate methods like removeAllFromXXXCollection, withXXXField as there is also a removeAllFromCollection, withField, etc.

### DIFF
--- a/annotations/builder/src/main/java/io/sundr/builder/internal/functions/ClazzAs.java
+++ b/annotations/builder/src/main/java/io/sundr/builder/internal/functions/ClazzAs.java
@@ -23,6 +23,7 @@ import static io.sundr.model.utils.Types.isAbstract;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.HashSet;
+import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
@@ -50,6 +51,7 @@ import io.sundr.model.ClassRef;
 import io.sundr.model.ClassRefBuilder;
 import io.sundr.model.Method;
 import io.sundr.model.MethodBuilder;
+import io.sundr.model.MethodFluent.BlockNested;
 import io.sundr.model.Modifiers;
 import io.sundr.model.Property;
 import io.sundr.model.PropertyBuilder;
@@ -112,6 +114,8 @@ public class ClazzAs {
       constructors.add(emptyConstructor);
       constructors.add(instanceConstructor);
 
+      Set<Property> allDescendants = new LinkedHashSet<>();
+
       item.getAllProperties().stream().filter(isPropertyApplicable(item)).forEach(property -> {
         final TypeRef unwrapped = TypeAs.combine(TypeAs.UNWRAP_ARRAY_OF, TypeAs.UNWRAP_COLLECTION_OF, TypeAs.UNWRAP_OPTIONAL_OF)
             .apply(property.getTypeRef());
@@ -137,6 +141,7 @@ public class ClazzAs {
             .withComments().withAnnotations().build();
 
         Set<Property> descendants = Descendants.PROPERTY_BUILDABLE_DESCENDANTS.apply(toAdd);
+        allDescendants.addAll(descendants);
         toAdd = new PropertyBuilder(toAdd).addToAttributes(DESCENDANTS, descendants).accept(new InitEnricher()).build();
         List<Method> methods = new ArrayList<Method>();
         if (isArray) {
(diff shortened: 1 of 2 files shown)


shawkins:
@iocanel do you have any thoughts on this direction - moving towards removing what could be seen as extra collection methods?

annotations/builder/src/main/java/io/sundr/builder/internal/functions/ClazzAs.java
```diff
@@ -254,6 +270,37 @@ public TypeDef apply(RichTypeDef item) {
                   .accept(new AddNoArgWithMethod())
                   .build());
     }
+
+    private Method createDescendantBuilderMethod(Set<Property> allDescendants) {
+      BlockNested<MethodBuilder> builderBuilder = new MethodBuilder().withNewModifiers().withProtected().withStatic()
+          .endModifiers()
+          .withParameters(Types.T)
+          .withArguments(new PropertyBuilder().withName("item").withTypeRef(Types.OBJECT_REF).build())
```
iocanel, line 278: You can use `addNewArgument` here to avoid manually creating the property builder.

annotations/builder/src/main/java/io/sundr/builder/internal/functions/ClazzAs.java
```diff
@@ -254,6 +270,37 @@ public TypeDef apply(RichTypeDef item) {
...
+      for (Property descendant : allDescendants) {
+        ClassRef dunwraped = new ClassRefBuilder(
+            (ClassRef) TypeAs.combine(TypeAs.UNWRAP_COLLECTION_OF, TypeAs.UNWRAP_ARRAY_OF, TypeAs.UNWRAP_OPTIONAL_OF)
+                .apply(descendant.getTypeRef())).withArguments(new TypeRef[0]).build();
+        String className = dunwraped.getFullyQualifiedName();
+        if (!seen.add(className) || !isBuildable(dunwraped)) {
+          continue;
+        }
+        // use a string concat for the case key to prevent conversion of fully qualified names to short names 
+        String packageName = dunwraped.getPackageName();
```
iocanel, line 292: Why not `dunwrapped.getFullyQualifiedName()`?

iocanel:
Just to make sure we are on the same page. The idea is that by adding a method that creates the builder of a class, we don't need a concrete impl per descendant, right?

This sounds like a good idea to me. 

I think that instead of deprecating these methods we should completely remove them. 
Next release (0.100.0) is going to have big incompatible changes anyway and our major version number does not promise stability anyway :smile: .

Regarding the idea of using a lookup table vs reflection, IMHO it makes sense if we can completely eliminate reflection. I wouldn't add it to this PR though as its already long and hard to fully review.

shawkins:
> Just to make sure we are on the same page. The idea is that by adding a method that creates the builder of a class, we don't need a concrete impl per descendant, right?

Yes, the only difference between removeAllFromSub1Collection and removeAllFromSub2Collection is that creation of the respective builders.  There is already, or one is added by these changes, a more general method removeAllFromCollection that accepts the super type and will use the common builder method.

> I think that instead of deprecating these methods we should completely remove them.

Ok, I'll update the pr with that.

> Regarding the idea of using a lookup table vs reflection, IMHO it makes sense if we can completely eliminate reflection. I wouldn't add it to this PR though as its already long and hard to fully review.

The lookup table/switch is part of this pr - createDescendantBuilderMethod.  In fabric8 we have if/else instanceOf chains that are 250 types long, so it seems advantageous to use a switch.  If the item matches one of those classes directly, then we'll create the corresponding builder.  If it's editable in general, in particular the automatically created EditableFoo classes, then the fall through call to builderOf will handle it.  So only if item class is not in directly in the descendant list and the item is not editable would builderOf then use reflection - are there cases where users create subtypes of non-editable pojos, but still expect them to use the builder of the parent type?

shawkins:
@iocanel is there a snapshot available? We can move fabric8 master to that to resolve the downstream check error.

Issue #356, opened 2023-01-03
Invalid generated code


andreaTP:
Steps to reproduce:

- write a file with the following content: [istio.csv](https://github.com/sundrio/sundrio/files/10339129/istio.csv) and rename it to `istio.yaml`
- `jbang io.fabric8:java-generator-cli:6.3.1 --add-extra-annotations --source istio.yaml --target ./src`
- write a file `istio.java` with the following content:
```
//DEPS io.fabric8:kubernetes-client:6.3.1
//DEPS io.fabric8:generator-annotations:6.3.1
//DEPS io.sundr:builder-annotations:0.93.2
//DEPS org.projectlombok:lombok:1.18.24
//SOURCES src/**.java
```
- `jbang build --fresh istio.java`

Produces the following compilation errors:

```
/Users/aperuffo/.jbang/cache/jars/istio.java.59e8d5798babedaffa7fe6fcb722e97356eece8d00b8f6b2130c9b39169e5d22.jar.tmp/io/istio/networking/v1beta1/virtualservicespec/HttpFluentImpl.java:465: error: method getMirrorPercent() is already defined in class HttpFluentImpl
  public Long getMirrorPercent() {
              ^
/Users/aperuffo/.jbang/cache/jars/istio.java.59e8d5798babedaffa7fe6fcb722e97356eece8d00b8f6b2130c9b39169e5d22.jar.tmp/io/istio/networking/v1beta1/virtualservicespec/HttpFluentImpl.java:468: error: method withMirrorPercent(Long) is already defined in class HttpFluentImpl
  public A withMirrorPercent(Long mirror_percent) {
           ^
  where A is a type-variable:
    A extends HttpFluent<A> declared in class HttpFluentImpl
/Users/aperuffo/.jbang/cache/jars/istio.java.59e8d5798babedaffa7fe6fcb722e97356eece8d00b8f6b2130c9b39169e5d22.jar.tmp/io/istio/networking/v1beta1/virtualservicespec/HttpFluentImpl.java:471: error: method hasMirrorPercent() is already defined in class HttpFluentImpl
  public Boolean hasMirrorPercent() {
                 ^
[...]

PR #382, opened 2023-05-29
Fail fast when creating builder that have dublicating properties after camel case convertion.

iocanel:
Resolves: #356


andreaTP (approved):
LGTM thanks!

Issue #384, opened 2023-06-01
Breaking changes for 0.100.0 release


shawkins:
Just wanted to catalog some possible breaking changes for 0.100.0, which may be eventually moved into the readme.md

* remove deprecated getXXX methods, which should use buildXXX instead
* from #379 removal of methods that reference a concrete descendant type that remove/add/removeMatching from a collection or use with.  Instead a general method should be used - removeAllFromCollection(SuperType... value).  This will be a hard breaking change in that the methods will not have been previously deprecated, which may mean downstream projects may want to avoid this version change in a fix release.

shawkins:
@iocanel is there anywhere else you want this captured?

iocanel:
> @iocanel is there anywhere else you want this captured?
Don't have something to add.

shawkins:
Marking as resolved.  Added a changelog based upon this for https://github.com/fabric8io/kubernetes-client/pull/5308

PR #389, opened 2023-06-12
Add isEnumConstant and isSynthetic to the Property model

andreaTP:
@iocanel I need a little assistance here, as re-generating the model Builders and Fluent with `(cd model && make)` doesn't produce the expected result.

The intention here is to surface the information easily accessible using the reflection API, this change should enable a proper fix for:
https://github.com/fabric8io/kubernetes-client/issues/5228
https://github.com/fabric8io/kubernetes-client/issues/4225

### DIFF
--- a/adapters/reflect/src/main/java/io/sundr/reflect/ClassTo.java
+++ b/adapters/reflect/src/main/java/io/sundr/reflect/ClassTo.java
@@ -278,9 +278,12 @@ public class ClassTo {
             .filter(c -> !item.equals(c))
             .collect(Collectors.toList()));
       }
+
       properties.add(new PropertyBuilder()
           .withName(field.getName())
           .withModifiers(Modifiers.from(field.getModifiers()))
+          .isEnumConstant(field.isEnumConstant())
+          .isSynthetic(field.isSynthetic())
           .withAnnotations(annotationRefs)
           .withTypeRef(TYPEREF.apply(field.getGenericType()))
           .build());
--- a/adapters/reflect/src/main/java/io/sundr/reflect/ClassTo.java
+++ b/adapters/reflect/src/main/java/io/sundr/reflect/ClassTo.java
@@ -282,8 +282,8 @@ public class ClassTo {
       properties.add(new PropertyBuilder()
           .withName(field.getName())
           .withModifiers(Modifiers.from(field.getModifiers()))
-          .isEnumConstant(field.isEnumConstant())
-          .isSynthetic(field.isSynthetic())
+          .withEnumConstant(field.isEnumConstant())
+          .withSynthetic(field.isSynthetic())
           .withAnnotations(annotationRefs)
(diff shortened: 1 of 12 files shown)


adapters/reflect/src/main/java/io/sundr/reflect/ClassTo.java
```diff
@@ -278,9 +278,12 @@ private static Set<Property> getProperties(Class item, Set<Class> references) {
             .filter(c -> !item.equals(c))
             .collect(Collectors.toList()));
       }
+
       properties.add(new PropertyBuilder()
           .withName(field.getName())
           .withModifiers(Modifiers.from(field.getModifiers()))
+          .isEnumConstant(field.isEnumConstant())
+          .isSynthetic(field.isSynthetic())
```
iocanel, line 286:
```suggestion
          .withEnumConstant(field.isEnumConstant())
          .withSynthetic(field.isSynthetic())
```

model/base/src/main/java/io/sundr/model/Property.java
```diff
@@ -32,18 +32,23 @@ public class Property extends ModifierSupport implements Renderable, Commentable
   private final TypeRef typeRef;
   private final String name;
   private final List<String> comments;
+  private final boolean enumConstant;
+  private final boolean synthetic;
 
-  public Property(List<AnnotationRef> annotations, TypeRef typeRef, String name, List<String> comments, Modifiers modifiers,
```
iocanel, line 36: I think that you should keep the old constructor around for compatibility reasons.
Make sure the old constructor is after the new so that builders get generated with the new one (I think the first non empty is picked or something like that).

iocanel (commented):
I am not exactly sure what issue you had, but here's my experience with your PR:

Upgrading the model with `make` worked fine for me in following builds the generator was broken. I think that there are two ways we could go about. 

1. Just add the two fields on `Property` check in the newly generated classes, release and then aling everything to the new model before using the changes.

2. Add a second constructor in `Property` so that we maintain compatibility.

model/base/src/main/java/io/sundr/model/Property.java
andreaTP: done, thanks for the suggestions!

andreaTP:
@iocanel this is now ready for a final review, thanks for your comment you pointed me in the right direction.

andreaTP:
@iocanel thanks a lot! Would you be able to follow up with a release?

iocanel:
We can. The question is do you need a 0.100.x or a 0.9x?

andreaTP:
Those changes are not breaking, no strong preference, I let @shawkins decide.

Thanks a lot for getting back!

shawkins:
These are not, but other changes in made could be considered breaking - #384

andreaTP:
0.100 seems to be the answer! 🎉

PR #391, opened 2023-06-20
calling the other constructor

shawkins:
Just a small change to call the other constructor, rather than repeating the logic.


shawkins:
Arg, made this change too quickly.  this can't be referenced yet.

PR #392, opened 2023-06-21
chore: prepare for sundrio 0.100.0

iocanel:
cc @shawkins @andreaTP


andreaTP:
Thanks a lot @iocanel !

PR #395, opened 2023-06-23
fix #394: ensures that defaults are copied

### LINKED ISSUE #394: The change to accept a null instance is not preserving defaults
#357 in combination with changes to the nested builder constructors from #385 causes the use of nested withNew... methods to not set default values.

shawkins:


iocanel (approved):
lgtm

Issue #398, opened 2023-06-28
Collections.extractArgument fails on non-parameterised class


metacosm:
`Map`, `Map<?,?>` fields result in:
```java
Caused by: java.lang.IndexOutOfBoundsException: Index 0 out of bounds for length 0
    at jdk.internal.util.Preconditions.outOfBounds (Preconditions.java:64)
    at jdk.internal.util.Preconditions.outOfBoundsCheckIndex (Preconditions.java:70)
    at jdk.internal.util.Preconditions.checkIndex (Preconditions.java:248)
    at java.util.Objects.checkIndex (Objects.java:372)
    at java.util.ArrayList.get (ArrayList.java:459)
    at io.sundr.model.utils.Collections.lambda$extractArgument$0 (Collections.java:208)
    at java.util.Optional.map (Optional.java:265)
    at io.sundr.model.utils.Collections.extractArgument (Collections.java:208)
```

metacosm:
Hmm, this doesn't seem to occur all the time but seems to occur when the type with the unparameterised field is part of a rather multi-level inheritance hierarchy.

iocanel:
> `Map`, `Map<?,?>` fields result in:

It's not clear to me under which conditions this is causing issues.

So, the only thing I can do for now is add a bounds check in `Collections.extractArgument`

PR #399, opened 2023-07-03
fix: check arguments size before calling get

iocanel:
Resolves: #398


iocanel:
@metacosm: does this fix your issue?

PR #402, opened 2023-07-03
fix #394 instance needs to be fully qualified and not abstract

### LINKED ISSUE #394: The change to accept a null instance is not preserving defaults

shawkins:
@iocanel sorry about this, ran through the snapshot again with fabric8 and saw a situation where the instance type is fully qualified and the type with the same name is not, which led to a compilation error.  So this needs to be fully qualified.  A simple reproducer didn't seem to induce the behavior though.


shawkins:
After this I was able to successfully align fabric8 to a new sundrio.

iocanel (approved):
lgtm

PR #403, opened 2023-07-04
fix #393: detecting all settable properties

### LINKED ISSUE #393: initialization behavior with MutablePersonWrapperBuilder
This class looks like:

```
public MutablePersonWrapperBuilder(MutablePersonWrapper instance,Boolean validationEnabled) {
    this.fluent = this; 
    if (instance != null) {
    }
    this.validationEnabled = validationEnabled; 
  }
  MutablePersonWrapperFluent<?> fluent;
  Boolean validationEnabled;
  public EditableMutablePersonWrapper build() {
    EditableMutablePersonWrapper buildable = new EditableMutablePersonWrapper();
    buildable.setItem(fluent.buildItem());
    return buildable;
  }
```
There's no attempt to initialize the builder item when the instance is not null.  This behavior is not new, in .94 it looked like:
```
  public MutablePersonWrapperBuilder(MutablePersonWrapper instance,Boolean validationEnabled) {
    this.fluent = this; 
    this.validationEnabled = validationEnabled; 
  }
```
Shouldn't there be an attempt to call withItem(instance.getItem())?

shawkins:
Uses allProperties similar to the build logic to find all fields that can be set.

cc @metacosm this should fix CustomResource subclass builder initialization, such as https://github.com/keycloak/keycloak/issues/20872

### DIFF
--- a/annotations/builder/src/main/java/io/sundr/builder/internal/functions/ClazzAs.java
+++ b/annotations/builder/src/main/java/io/sundr/builder/internal/functions/ClazzAs.java
@@ -609,7 +609,7 @@ public class ClazzAs {
 
   public static final Function<RichTypeDef, TypeDef> POJO = FunctionFactory.wrap(new ToPojo());
 
-  private static List<Statement> toInstanceConstructorBody(TypeDef clazz, TypeDef instance, String fluent) {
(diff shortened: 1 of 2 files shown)


iocanel (approved):
lgtm

PR #408, opened 2023-07-31
fix #407: sort method names before finding a match

### LINKED ISSUE #407: Fluent Builder code generation issue
When an object has the following class definition:

```
public class IstioDestinationRule implements KubernetesResource
{

    @JsonProperty("canarySubsetName")
    private String canarySubsetName;
    @JsonProperty("name")
    private String name;
    @JsonProperty("stableSubsetName")
    private String stableSubsetName;
    @JsonIgnore
    private Map<String, Object> additionalProperties = new LinkedHashMap<String, Object>();

    @JsonProperty("canarySubsetName")
    public String getCanarySubsetName() {
        return canarySubsetName;
    }

    @JsonProperty("canarySubsetName")
    public void setCanarySubsetName(String canarySubsetName) {
        this.canarySubsetName = canarySubsetName;
    }

    @JsonProperty("name")
    public String getName() {
        return name;
    }

    @JsonProperty("name")
    public void setName(String name) {
        this.name = name;
    }
```

When using the Fluent builder annotation processing, there exists an issue with matching the correct setter function when calling `Setter.find` here: https://github.com/sundrio/sundrio/blob/main/annotations/builder/src/main/java/io/sundr/builder/internal/functions/ClazzAs.java#L679

Due to the way the `Setter.find` logic works, the property `name` gets matched with `setCanarySubsetName` since the function ends with `setName`.

ryantse:
Modifies the `Setter.find` functionality to first sort the method names by length prior to trying to find a match for a particular property name.

Addresses an edge case in finding a setter where there may be properties such as `canarySubsetName` while also having a `name` property which may cause unintended matches when trying to find the `setName` setter for the `name` property.

### DIFF
--- a/model/utils/src/main/java/io/sundr/model/utils/Setter.java
+++ b/model/utils/src/main/java/io/sundr/model/utils/Setter.java
@@ -18,6 +18,10 @@ package io.sundr.model.utils;
 
 import static io.sundr.utils.Strings.capitalizeFirst;
 
+import java.util.Comparator;
+import java.util.List;
+import java.util.stream.Collectors;
+
 import io.sundr.SundrException;
 import io.sundr.model.Method;
 import io.sundr.model.Property;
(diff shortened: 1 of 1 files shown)


ryantse:
@iocanel: Would you be able to help review this change for fixing this issue?

iocanel (approved):
The approach is really smart and I like it.
It would be great if we could avoid sorting, but for now its good.

PR #416, opened 2023-09-09
Removes validationEnabled constructors when not in use

### LINKED ISSUE #415: Remove unneeded builder constructors when validation is not enabled
From https://github.com/fabric8io/kubernetes-client/issues/5409 there are extraneous builder constructors that could be removed.

shawkins:
also consolidates instance copying

Closes #415

@iocanel in looking at testing can you clarify what the expectations are for the feature - in classAs we check both at a type level and at a BuildContext level for validationEnabled.  However the BuildContextManager will not adjust the value of validationEnabled on the context.  Are users supposed to mark all types as validationEnabled=true?


shawkins:
Took a guess at how the context logic should work by allowing it to accumulate any true setting of validationEnabled.

shawkins:
> @shawkins: It seems that the changes in constructors has a slight impact it one of the client extension examples.
Is this something we can address?

Yes, those aren't needed: https://github.com/fabric8io/kubernetes-client/pull/5440

iocanel (approved):
LGTM

manusa (approved):
Thx

shawkins:
@iocanel The upstream has been updated, this can be retested.

PR #431, opened 2023-10-20
simplifications to the typing of basefluent

shawkins:
fixes fabric8io/kubernetes-client#5522


shawkins:
@iocanel the core of the issue is that implementing Visitable twice is confusing the compiler, this removes that but also does a few other things:
- removes the Fluent interface, we don't seem to need it any more
- simplifies BaseFluent to just BaseFluent<F> - we don't really need a bound on this because we only ever use it via SomethingFluent, we could even make it abstract.
- adds a getVisitableMap() on Visitables so that we can pushdown all of the visitable logic.  However I took a shortcut and put the implementation of this method on BaseFluent.

Also I'm assuming that SomethingFluent does not actually need to be Visitable, but I could be wrong about that.

shawkins:
@iocanel are the any changes that you want on this pr?  We'd like to get it into a fabric8 release so that the operator sdk can upgrade without changing their code.

annotations/builder/src/main/java/io/sundr/builder/internal/BuilderContext.java
```diff
@@ -1077,66 +1074,8 @@ public BuilderContext(Elements elements, Types types, Boolean generateBuilderPac
 
         .addNewMethod()
         .withNewModifiers().withPublic().endModifiers()
-        .withName("accept")
```
iocanel, line 1080: While we do drop this large chunk of boilerplate and I am grateful for it, I cannot help wonder what exactly allow us to do so.

I understand that we moved the target methods from BaseFluent to Visitable, but don't we need to move their definition too? In other words  should we construct a Visitable TypeDef that includes these `accept` methods?

iocanel:
@shawkins I did add a question.

annotations/builder/src/main/java/io/sundr/builder/internal/BuilderContext.java
shawkins: The Visitable TypeDef already has the necessary logic to support the accept default methods.  The only new method is getVisitableMap, which handling was added for on both Visitable and BaseFluent.  The one bit of messiness there is that it has to be added to Visitable after VisitableMap is defined - is there a better way to do that?

iocanel:
BTW @shawkins: excellent job. I salute!

shawkins:
> BTW @shawkins: excellent job. I salute!

Thank you @iocanel - it was good to learn more of the code, you have a humbling amount of work here.

PR #440, opened 2023-10-27
fix: remove the named field, not the instance

### LINKED ISSUE #438: Visitable.accept visits all Visitor instances regardless of the type
Changes in #431 seem to be causing an issue with Visitors

With the new changes in:

https://github.com/sundrio/sundrio/blob/836ebeabed8f708f792ffac001a43e154852f0ea/core/src/main/java/io/sundr/builder/Visitable.java#L77-L91

Visitors are now called regardless of the matching type.

Seems like the double for-loop is a mistake or something left from previous iterations.

/cc @shawkins @iocanel

shawkins:
closes #438


iocanel (commented):
Any chance to include a test, so that we prevent future regressions?

shawkins:
@manusa mentioned he was adding one https://github.com/sundrio/sundrio/issues/438#issuecomment-1782943188 - if not, I can add something.

The interesting thing here is that this logic has been in place for a while, but wasn't coming into play with the straight-forward building example that Marc's using.  I'll try to compare what was happening is older releases a little more closely.

manusa:
I'm sorry, I couldn't find the way to reproduce here. 
This is my WIP [FeatureBuilder](https://github.com/sundrio/sundrio/pull/441), I'm sure I'm missing something but I can't see it.

shawkins:
>  I'll try to compare what was happening is older releases a little more closely.

The other part of this issue is that withXXX is being called twice in the copyInstance method.  I'm see about adding a fix for that as well.

shawkins:
@manusa @iocanel this has been updated with a more complete fix and a direct test of the duplidate Visitable behavior.

iocanel (approved):
LGTM

PR #445, opened 2023-11-06
fix: limiting when prefixing is used

### LINKED ISSUE #437: [Builder Annotation] : Builder methods generated `withFooFoo` for fields with same type name but different packages
Follow up to #425

I upgraded to newer sundrio version and I can no longer see the error. Thanks.

However, I see some additional changes in builder methods which I did not expect:
```diff
--- a/kubernetes-model-generator/openshift-model-hive/src/test/java/io/fabric8/openshift/api/model/hive/v1/ClusterPoolTest.java
+++ b/kubernetes-model-generator/openshift-model-hive/src/test/java/io/fabric8/openshift/api/model/hive/v1/ClusterPoolTest.java
@@ -67,12 +67,12 @@ class ClusterPoolTest {
         .withName("openshift-4-6")
         .endImageSetRef()
         .withNewPlatform()
-        .withNewAws()
+        .withNewAwsAws()
         .withNewCredentialsSecretRef()
         .withName("hive-team-aws-creds")
         .endCredentialsSecretRef()
         .withRegion("us-east-1")
-        .endAws()
+        .endAwsAws()
         .endPlatform()
         .withNewPullSecretRef()
         .withName("hive-team-pull-secret")
```

Shall I create a new issue about this? You can inspect the changes  affected in this commit [4db1082d4e670c3cca69b51ed5b69439ece86afc](https://github.com/rohanKanojia/kubernetes-client/commit/4db1082d4e670c3cca69b51ed5b69439ece86afc)

_Originally posted by @rohanKanojia in https://github.com/sundrio/sundrio/issues/425#issuecomment-1778465943_

shawkins:
Simple buildable properties with the change are not prefixed as they don't need additional package qualification.

closes #437

### DIFF
--- a/annotations/builder/src/main/java/io/sundr/builder/internal/functions/PropertyAs.java
+++ b/annotations/builder/src/main/java/io/sundr/builder/internal/functions/PropertyAs.java
@@ -213,10 +213,15 @@ public final class PropertyAs {
       }
       parameters.add(N);
 
+      // synthetic properties of descendants need a prefix as the class names may collide
+      String prefix = property.hasAttribute(Constants.DESCENDANT_OF)
+          ? BuilderUtils.fullyQualifiedNameDiff(property.getTypeRef(), originTypeDef)
+          : "";
+
       return new TypeDefBuilder()
           .withKind(Kind.INTERFACE)
           .withPackageName(outerClass.getPackageName())
-          .withName(BuilderUtils.fullyQualifiedNameDiff(property.getTypeRef(), originTypeDef) + property.getNameCapitalized()
+          .withName(prefix + property.getNameCapitalized()
               + "Nested")
           // all references are local - qualified causes compilation errors
           //.withOuterTypeName(outerClass.getFullyQualifiedName())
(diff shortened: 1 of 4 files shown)


annotations/builder/src/main/java/io/sundr/builder/internal/functions/PropertyAs.java
```diff
@@ -213,10 +213,14 @@ public TypeDef apply(Property property) {
       }
       parameters.add(N);
 
+      String prefix = property.hasAttribute(Constants.DESCENDANT_OF)
```
iocanel, line 216: I am not sure I understand the logic here.
Why are we adding the diff only when we have an ancestor property ? 
Naming conflicts are not limited there.

iocanel (commented):
The reasoning is not clear to me. Can you please elaborate why we are limiting the scope to ancestor properties?

annotations/builder/src/main/java/io/sundr/builder/internal/functions/PropertyAs.java
shawkins: My understanding is the logic creates synthetic properties for descendents, that's where we want to use the prefixing as the subtypes may collide.  For something that is just a buildable property without multiple subtypes, no prefix is needed - the property name will already be unique.

iocanel:
> > The reasoning is not clear to me. Can you please elaborate why we are limiting the scope to ancestor properties?
> 
> @iocanel is there anything else you'd like to see as an explaination?

The explanation is fine. A comment on the code would be really nice to have.

iocanel (approved):
LGTM

PR #446, opened 2023-11-06
fix: allows for class names to not start with upper case letters

### LINKED ISSUE #430: Issues in generation when a type starts with special characters
Issue found in the `camel-k` project:
https://github.com/apache/camel-k/blob/f2d3fd3b24aa2991940dd3f729074914a0c948b5/config/crd/bases/camel.apache.org_integrationplatforms.yaml#L437

In case a class name starts with `_` (as the first character has escaped e.g. it was a number) the generation fails here:
https://github.com/sundrio/sundrio/blob/371398cbb6ed0e330670da87a9e53950de769c89/model/base/src/main/java/io/sundr/model/Nameable.java#L28

Unfortunately, just adding a check there doesn't completely fix the problem.

This is easily reproducible with a `pom.xml`:

```xml
<?xml version="1.0" encoding="UTF-8"?>
<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">

  <parent>
    <groupId>org.apache</groupId>
    <artifactId>apache</artifactId>
    <version>23</version>
  </parent>

  <modelVersion>4.0.0</modelVersion>
  <groupId>org.apache.camel.k</groupId>
  <artifactId>camel-k-crds</artifactId>
  <version>2.1.0-SNAPSHOT</version>
  <packaging>jar</packaging>

  <properties>
    <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>

    <maven.compiler.release>11</maven.compiler.release>
    <maven.compiler.source>11</maven.compiler.source>
    <maven.compiler.target>11</maven.compiler.target>

    <fabric8-version>6.9.0</fabric8-version>
    <lombok-version>1.18.30</lombok-version>
    <sundrio-version>0.101.0</sundrio-version>
  </properties>

  <dependencyManagement>
    <dependencies>
          <dependency>
              <groupId>io.fabric8</groupId>
              <artifactId>kubernetes-client-bom</artifactId>
              <version>${fabric8-version}</version>
              <type>pom</type>
              <scope>import</scope>
          </dependency>
      </dependencies>
  </dependencyManagement>

  <dependencies>
    <dependency>
      <groupId>io.fabric8</groupId>
      <artifactId>kubernetes-client</artifactId>
    </dependency>
    <dependency>
      <groupId>io.fabric8</groupId>
      <artifactId>generator-annotations</artifactId>
    </dependency>
    <dependency>
      <groupId>io.sundr</groupId>
      <artifactId>builder-annotations</artifactId>
      <version>${sundrio-version}</version>
      <scope>provided</scope>
    </dependency>
    <dependency>
      <groupId>org.projectlombok</groupId>
      <artifactId>lombok</artifactId>
      <version>${lombok-version}</version>
      <scope>provided</scope>
    </dependency>
  </dependencies>

  <build>
    <plugins>
      <plugin>
        <groupId>io.fabric8</groupId>
        <artifactId>java-generator-maven-plugin</artifactId>
        <version>${fabric8-version}</version>
        <executions>
          <execution>
            <goals>
              <goal>generate</goal>
            </goals>
          </execution>
        </executions>
        <configuration>
          <source>dummy-crd.yml</source>
          <extraAnnotations>true</extraAnnotations>
        </configuration>
      </plugin>
    </plugins>
  </build>

</project>
```

and a `dummy-crd.yml` like:

```yaml
apiVersion: apiextensions.k8s.io/v1
kind: CustomResourceDefinition
metadata:
  name: dummies.example.com
spec:
  group: example.com
  versions:
    - name: v1
      served: true
      storage: true
      schema:
        openAPIV3Schema:
          type: object
          properties:
            spec:
              type: object
              properties:
                3scale:
                  description: ''
                  properties:
                    configuration:
                      description: TraitConfiguration parameters configuration
                      type: object
                      x-kubernetes-preserve-unknown-fields: true
                  type: object
  scope: Namespaced
  names:
    plural: dummies
    singular: dummy
    kind: Dummy
```

running the command `mvn install` will trigger the reproducer.

shawkins:
@iocanel the only issue I can see here is that you can't have inner types if you have a class that starts with something other than an upper case.

@andreaTP you could consider changing the generator convention to prepending A instead of _ and everything would work as expected with sundrio.

closes #430

### DIFF
--- a/model/base/src/main/java/io/sundr/model/Nameable.java
+++ b/model/base/src/main/java/io/sundr/model/Nameable.java
@@ -66,13 +66,27 @@ public interface Nameable extends Node {
   }
 
   static String getClassName(String fullyQualifiedName) {
-    return Arrays.stream(fullyQualifiedName.split(PACKAGE_SEPARATOR_REGEX)).filter(after(IS_UPPER_CASE))
-        .collect(Collectors.joining(DOT));
+    String[] parts = fullyQualifiedName.split(PACKAGE_SEPARATOR_REGEX);
+
+    String result = Arrays.stream(parts).filter(after(IS_UPPER_CASE)).collect(Collectors.joining(DOT));
+
+    if (!result.isEmpty()) {
+      return result;
+    }
+
+    return parts[parts.length - 1];
   }
 
   static String getPackageName(String fullyQualifiedName) {
-    return Arrays.stream(fullyQualifiedName.split(PACKAGE_SEPARATOR_REGEX)).filter(until(IS_UPPER_CASE))
-        .collect(Collectors.joining(DOT));
+    String[] parts = fullyQualifiedName.split(PACKAGE_SEPARATOR_REGEX);
+
+    String result = Arrays.stream(parts).filter(until(IS_UPPER_CASE)).collect(Collectors.joining(DOT));
+
+    if (!result.equals(fullyQualifiedName)) {
+      return result;
+    }
+
+    return Arrays.asList(parts).subList(0, parts.length - 1).stream().collect(Collectors.joining(DOT));
   }
 
   static <T> Predicate<T> until(final Predicate<T> predicate) {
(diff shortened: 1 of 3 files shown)


iocanel (approved):
LGTM

andreaTP:
Thanks a lot for taking care of this @shawkins 👍

Issue #457, opened 2023-12-11
Unable to build project with maven 3.9.6


rohanKanojia:
I was trying to test changes in https://github.com/sundrio/sundrio/pull/455 but I noticed I'm not able to build project with maven 3.9. Build seems to fail in sundrio core due to some NoClassDefFoundError error. 

Not 100% sure but it might be due to the impsort-maven-plugin not being compatible with maven 3.9. 

```
[ERROR] Failed to execute goal net.revelc.code:impsort-maven-plugin:1.4.1:sort (default) on project sundr-core: Execution default of goal net.revelc.code:impsort-maven-plugin:1.4.1:sort failed: A required class was missing while executing net.revelc.code:impsort-maven-plugin:1.4.1:sort: null
[ERROR] -----------------------------------------------------
[ERROR] realm =    plugin>net.revelc.code:impsort-maven-plugin:1.4.1
[ERROR] strategy = org.codehaus.plexus.classworlds.strategy.SelfFirstStrategy
[ERROR] urls[0] = file:/home/rokumar/.m2/repository/net/revelc/code/impsort-maven-plugin/1.4.1/impsort-maven-plugin-1.4.1.jar
[ERROR] urls[1] = file:/home/rokumar/.m2/repository/com/github/javaparser/javaparser-core/3.15.21/javaparser-core-3.15.21.jar
[ERROR] Number of foreign imports: 1
[ERROR] import: Entry[import  from realm ClassRealm[maven.api, parent: null]]
[ERROR] 
[ERROR] -----------------------------------------------------
[ERROR] : NoClassDefFoundError: org/codehaus/plexus/util/DirectoryScanner
[ERROR] -> [Help 1]
[ERROR] 
[ERROR] To see the full stack trace of the errors, re-run Maven with the -e switch.
[ERROR] Re-run Maven using the -X switch to enable full debug logging.
[ERROR] 
[ERROR] For more information about the errors and possible solutions, please read the following articles:
[ERROR] [Help 1] http://cwiki.apache.org/confluence/display/MAVEN/PluginContainerException
[ERROR] 
[ERROR] After correcting the problems, you can resume the build with the command
[ERROR]   mvn <args> -rf :sundr-core
```

This doesn't happen with maven wrapper. I can build the project successfully when I use maven wrapper.

My maven version details:
```
Apache Maven 3.9.6 (bc0240f3c744dd6b6ec2920b3cd08dcc295161ae)
Maven home: /opt/apache-maven
[...]

PR #459, opened 2023-12-11
fix: upgrade mvn / mvnw support to 3.9

shawkins:
closes: #457


shawkins:
cc @rohanKanojia

rohanKanojia:
@shawkins : Thanks a lot for your fix! I tested it on JDK11 + JDK17 and I don't see the error anymore. I also reviewed the changes, apart from plugin version upgrades most changes are for removing trailing whitespace (probably added via plugin).

However, I'm not able to build project on JDK8 . I think there is some configuration mistake on my side. I'm getting this error when compiling on JDK8:
```
[ERROR] Failed to execute goal net.revelc.code.formatter:formatter-maven-plugin:2.23.0:format (default) on project sundr-pom: Execution default of goal net.revelc.code.formatter:formatter-maven-plugin:2.23.0:format failed: Unable to load the mojo 'format' in the plugin 'net.revelc.code.formatter:formatter-maven-plugin:2.23.0' due to an API incompatibility: org.codehaus.plexus.component.repository.exception.ComponentLookupException: net/revelc/code/formatter/FormatterMojo has been compiled by a more recent version of the Java Runtime (class file version 55.0), this version of the Java Runtime only recognizes class file versions up to 52.0
[ERROR] -----------------------------------------------------
[ERROR] realm =    plugin>net.revelc.code.formatter:formatter-maven-plugin:2.23.0
[ERROR] strategy = org.codehaus.plexus.classworlds.strategy.SelfFirstStrategy

```

shawkins:
@rohanKanojia I think what is happening is that formatting is turned off for ci runs, so we don't see a problem there with java 8. Do you think we should try to make formatting still work with java 8?

rohanKanojia:
> Do you think we should try to make formatting still work with java 8?

@shawkins : In my opinion, it's not that important.

iocanel (approved):
LGTM

Issue #466, opened 2024-04-02
Baseline To Java 11


shawkins:
Based upon the discusion on #464, support for Java 8 should be dropped.

_Originally posted by @iocanel in https://github.com/sundrio/sundrio/issues/464#issuecomment-2025639488_

PR #467, opened 2024-04-02
task: making the baseline java 11

shawkins:
closes: #466

Added an enforcer that the java version must be 11+  - that can be removed if it doesn't seem necessary.

maven-plugin-plugin needed to be updated as well, so I did a little alignment / update of the maven related dependencies - but didn't try to push it all the way to the latest.

removed any overrides that didn't seem to be needed given that they are coming from the parent.

had to add some compiler args to directly reference the com.sun stuff in the dsl test.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -29,7 +29,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        java: [8,11,17,21]
+        java: [11,17,21]
         profile: [normal, with-swagger]
     steps:
       - name: Checkout
@@ -42,7 +42,4 @@ jobs:
           cache: 'maven'
       - name: Build Project
         run: |
-          if [ -f $JAVA_HOME/lib/tools.jar ]; then
-            ./mvnw install:install-file -Dfile=$JAVA_HOME/lib/tools.jar -DgroupId=com.sun -DartifactId=tools -Dversion=8 -Dpackaging=jar -P${{ matrix.profile }};
-          fi
           ./mvnw ${MAVEN_ARGS} clean install -Pwith-examples -Dformat.skip=true
--- a/.github/workflows/downstream-check.yml
+++ b/.github/workflows/downstream-check.yml
@@ -32,7 +32,7 @@ jobs:
     if: ${{ github.event_name == 'pull_request' && contains(github.event.pull_request.labels.*.name, 'downstream-check') || github.event_name == 'push' && github.ref == 'refs/heads/main' }}
     strategy:
       matrix:
-        java: [8]
+        java: [11]
         profile: [normal]
     steps:
       - name: Checkout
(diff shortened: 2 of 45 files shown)


andreaTP (approved):
Lgtm! Thanks for taking this!

Have we decided to drop Java 8 support from the Kubernetes client too?

Otherwise we are going to be stuck on sundrio version ...

shawkins:
> Have we decided to drop Java 8 support from the Kubernetes client too?

Sounds good to me. 

I haven't tested, but I'm pretty sure you could keep using this java 11 version of sundrio to perform a java 8 compatible build of fabric8 as long as you are using a later jdk to perform the build.

andreaTP:
Let's discuss this tomorrow :+1:

iocanel (commented):
My understanding was that kubernetes-client would drop support for java 8 too.
If not this might be risky as @andreaTP already mentioned.

Also, I will need to check what the official client is doing in that regard cause I wouldn't want to cause break them either.

So, I will change my review till we short these out.

iocanel:
Ok, seems that the official client is already on java 11, so we just have to align with fabric8

andreaTP:
> Ok, seems that the official client is already on java 11, so we just have to align with fabric8

I cannot confirm, unfortunately:

https://github.com/fabric8io/kubernetes-client/blob/0c80a66d10bf7096864246ce453413e0da4edbd2/.github/workflows/release.yaml#L64

shawkins:
> I cannot confirm, unfortunately:

@andreaTP he's referring to https://github.com/kubernetes-client/java which does default to java 11 in their pom.

shawkins:
@andreaTP @iocanel the only other consideration here is if we do expect java 8 runtime support for the sundrio annotations / api.builder - the source / target can be set there to stay at java 8 if needed.

iocanel:
> @andreaTP @iocanel the only other consideration here is if we do expect java 8 runtime support for the sundrio annotations / api.builder - the source / target can be set there to stay at java 8 if needed.

What do you mean?

shawkins:
> What do you mean?

Those are the only parts of sundrio that should be used at runtime and could remain 8 compatible if needed.

iocanel:
> > What do you mean?
> 
> Those are the only parts of sundrio that should be used at runtime and could remain 8 compatible if needed.

That's a good point, but how would a project that is build with java 8. would be able to use our annotation processors that require java 11?

shawkins:
> That's a good point, but how would a project that is build with java 8. would be able to use our annotation processors that require java 11?

It's the same as what I'm mentioning for fabric8 - as long as you build with a 11+ jdk, you can still target 8 compatibility.

manusa:
The agreement for Fabric8 Kubernetes Client is to set the baseline to Java 11 on our next major (v7).
The initial plan is to do a last minor release (either v6.12 or v6.13) and then set the main branch SNAPSHOT version to v7 and start working on it.

andreaTP (approved):
Agreed 👍 thanks for the patience @shawkins

shawkins:
I tried to also set maven.compiler.release to 11, but it complains that you can't do that and use the compiler exports.

shawkins:
@iocanel we should be ready to commit this with your approval

PR #472, opened 2024-04-23
Prevent AptContext being used via TypeLoojup outside of apt

iocanel:


manusa (approved):

shawkins:
> I'm concerned about the fact that the build of Fabric8 client is failing with this change.

That looks to have been introduced by an earlier change.

shawkins:
@metacosm this change https://github.com/sundrio/sundrio/commit/ea807f89c80bffdf6098dfa77335c538f1a78ae9 would have caused the issue with fabric8. That code was only added last week to fabric8, so it wasn't seen in prior downstream check runs. Essentially the fabric8 code is now wrong - the methods are not correctly balanced:

```
.withNewSelector()
...
.endV1beta1Selector()
```

There's no released version of sundrio yet for us to pick up that has this addressed.

shawkins (approved):
LGTM

PR #475, opened 2024-04-25
chore: prepare for 0.200.0

iocanel:
As we baseline to java 11 I am bumping to 0.200.0.


iocanel:
cc @aureamunoz @shawkins @manusa @andreaTP @metacosm

PR #478, opened 2024-05-08
WITH_OPTIONAL now uses Expression API

iocanel:


iocanel:
@shawkins: can you please have a look? It's pretty much what the title says and its part of the effort of replacing string concatenation with a proper API for code generation.

shawkins (approved):
@iocanel sorry for the delayed response, I'm in the middle of a move. It does seem to work as expected - the existing coverage looks to hit the replacements.

Issue #470, opened 2024-04-15
IllegalStateException: on multiple properties clashing after sanitization


andreaTP:
Reproducer.

With a `pom.xml`:

```xml
<?xml version="1.0" encoding="UTF-8"?>
<project xmlns="http://maven.apache.org/POM/4.0.0"
         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
    <modelVersion>4.0.0</modelVersion>

    <groupId>io.repro</groupId>
    <artifactId>repro-sundrio</artifactId>
    <version>0.0.1</version>

    <packaging>jar</packaging>

    <name>Repro</name>
    
    <properties>
        <maven.compiler.source>17</maven.compiler.source>
        <maven.compiler.target>17</maven.compiler.target>
        <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
        <fabric8.version>6.12.0</fabric8.version>
    </properties>

    <dependencyManagement>
        <dependencies>
            <dependency>
                <groupId>io.fabric8</groupId>
                <artifactId>kubernetes-client-bom-with-deps</artifactId>
                <version>${fabric8.version}</version>
                <type>pom</type>
                <scope>import</scope>
            </dependency>
        </dependencies>
    </dependencyManagement>

    <dependencies>
        <dependency>
            <groupId>io.fabric8</groupId>
            <artifactId>openshift-client</artifactId>
        </dependency>
        <dependency>
            <groupId>io.fabric8</groupId>
            <artifactId>kubernetes-model</artifactId>
        </dependency>
        <dependency>
            <groupId>io.fabric8</groupId>
            <artifactId>generator-annotations</artifactId>
        </dependency>
        <dependency>
            <groupId>io.sundr</groupId>
            <artifactId>builder-annotations</artifactId>
            <scope>provided</scope>
        </dependency>
        <dependency>
            <groupId>org.projectlombok</groupId>
            <artifactId>lombok</artifactId>
            <scope>provided</scope>
        </dependency>
    </dependencies>

    <build>
        <plugins>
            <plugin>
                <groupId>io.fabric8</groupId>
                <artifactId>java-generator-maven-plugin</artifactId>
                <version>${fabric8.version}</version>
                <executions>
                    <execution>
                        <phase>generate-sources</phase>
                        <goals>
[...]

PR #479, opened 2024-05-12
fix: refined duplicate property check

shawkins:
closes: #470

Simplifies the duplicate property check logic and excludes enums.

There are two things that make naming lossy - getNameCapitalized, and compact. Ideally there could be some controls exposed to the user for this.


andreaTP (approved):

iocanel (approved):
LGTM

shawkins:
Ensured the other usage of compact stayed the same and added a test highlighting how compact should preserve whatever is the root of the name - that doesn't come into play for #470 directly, just a corner case that we can proactively address.

Issue #170, opened 2020-09-07
conflict method name in Fluent


kentchenjh:
Try below snippet to generate class file.

```
@Buildable(editableEnabled = false, validationEnabled = true, generateBuilderPackage = true, builderPackage = "foo.bar")
public class Playground {

    List<String> word;
    List<Integer> words;

    public List<String> getWord() {
        return word;
    }

    public void setWord(List<String> word) {
        this.word = word;
    }

    public List<Integer> getWords() {
        return words;
    }

    public void setWords(List<Integer> words) {
        this.words = words;
    }
}
```

result:

```
error: method getWord(int) is already is already defined in interface PlaygroundFluent
error: method getFirstWord() is already is already defined in interface PlaygroundFluent
```

It seems a bug.

iocanel:
@kentchenjh: This is an a known issue, that I haven't managed to find time to solve. I'll do my best the next couple of weeks.

PR #482, opened 2024-09-02
test: (disabled)reproducer test for duplicate builder methods fluent

manusa:
This is a reproducer test for what's more or less described in #170
It's disabled at the pom level (integration test) so that it can be merged before tackling the issue.

This is a big problem for the new generator approach in the Kubernetes Client (https://github.com/fabric8io/kubernetes-client/issues/6130)
As reported in https://github.com/fabric8io/kubernetes-client/issues/6320

The kind of naming and class structure as defined in this tests generates both duplicate builder methods and nested classes.

In this case, the generated `RootFluent` will contain:

```java
// 60
  public ObjectSpecNested<A> withNewObjectSpec() {
    return new ObjectSpecNested(null);
  }
// 72
  public ObjectSpecNested<A> withNewObjectSpec() {
    return new ObjectSpecNested(null);
  }
// 177
  public class ObjectSpecNested<N> extends ObjectSpecFluent<ObjectSpecNested<N>> implements Nested<N>{ //...
// 193
  public class ObjectSpecNested<N> extends ObjectFluent<ObjectSpecNested<N>> implements Nested<N>{ //...
```

A slight change in the name of the classes (Object or ObjectSpec) prevents the issue from happening.

### DIFF
--- a/tests/buildable-duplicates/pom.xml
+++ b/tests/buildable-duplicates/pom.xml
@@ -1,0 +1,42 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+       Copyright 2018 The original authors.
+
+       Licensed under the Apache License, Version 2.0 (the "License");
+       you may not use this file except in compliance with the License.
+       You may obtain a copy of the License at
+
+           http://www.apache.org/licenses/LICENSE-2.0
+
+       Unless required by applicable law or agreed to in writing, software
+       distributed under the License is distributed on an "AS IS" BASIS,
+       WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+       See the License for the specific language governing permissions and
+       limitations under the License.
+  -->
+
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <parent>
+        <artifactId>tests</artifactId>
+        <groupId>io.sundr</groupId>
+        <version>999-SNAPSHOT</version>
+    </parent>
+    <modelVersion>4.0.0</modelVersion>
+
+    <groupId>io.sundr.tests</groupId>
+    <artifactId>buildalbe-duplicates</artifactId>
+    <name>Sundrio :: Tests :: Buildable Duplicates</name>
+
+    <dependencies>
+        <dependency>
+            <groupId>io.sundr</groupId>
+            <artifactId>builder-annotations</artifactId>
+            <version>${project.version}</version>
+        </dependency>
+        <dependency>
+            <groupId>junit</groupId>
+            <artifactId>junit</artifactId>
+            <scope>test</scope>
+        </dependency>
+    </dependencies>
+</project>
--- a/tests/buildable-duplicates/src/main/java/io/sundr/tests/Object.java
+++ b/tests/buildable-duplicates/src/main/java/io/sundr/tests/Object.java
@@ -1,0 +1,40 @@
+/**
+ * Copyright 2015 The original authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+**/
+
+package io.sundr.tests;
+
+import io.sundr.builder.annotations.Buildable;
+
+@Buildable(editableEnabled = false, validationEnabled = false, generateBuilderPackage = true, lazyCollectionInitEnabled = false, builderPackage = "io.sundr.tests.builder")
+public class Object implements Resource {
+
+  private final ObjectSpec spec;
+  private final Resource status;
+
+  public Object(ObjectSpec spec, Resource status) {
+    this.spec = spec;
+    this.status = status;
+  }
+
+  public ObjectSpec getSpec() {
+    return spec;
+  }
+
+  public Resource getStatus() {
+    return status;
+  }
+}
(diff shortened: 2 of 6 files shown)


iocanel:
@manusa  thank you so much for the reproducer.

So what's going on here ?

We have a field `Resource spec` where `Resource` is an interface with two implementations:
- `Object`
- `ObjectSpec`

In this case sundrio will search for `@Buildable` types implementing `Resource` and will find both of the implementations mentioned above. 

For each of the formentioned types it will create method that follow the pattern: `with<Impl Name><Field name>`. So, we end up with:
- `withObjectSpec`
- `withObjectSpecSpec` that is normalized to `withObjectSpec`

Thus we end up with a conflict. The conflict is not just in methods but also in nested classes and possibly getters etc.

### Possible solutions
For that particular problem (that won't cover  #170):

- Filter out duplicates when rendering (resulting classes will end up with missing methods, but will compile).
- Add flag to control normalization (Accept `withObjectSpecSpec`. Yikes).
- Add a separator string in between <Impl Name> and <FieldName> (e.g. `As` -> `withObjectSpecAsSpec` and `withObjectAsSpec`).

For #170 I think that we need to provide the ability to define aliases for fields that will be used by sundrio for naming purposes, as I fail to see one solution that could possible cover all cases.

manusa:
> So what's going on here ?
> [...]
> - withObjectSpecSpec that is normalized to withObjectSpec

I didn't have time to look for the root cause, but what you describe seems to be the most plausible explanation.

> For that particular problem (that won't cover https://github.com/sundrio/sundrio/issues/170):

Given your explanation, I don't think the problem is related to #170 at all.

> Add a separator string in between and (e.g. As -> withObjectSpecAsSpec and withObjectAsSpec).

This seems to me the most easy way to deal with the problem. I'd vote for this solution.

iocanel:
@manusa: Then I'll work towards the described solution.

manusa:
> @manusa: Then I'll work towards the described solution.

Awesome, thanks!!! :raised_hands:

shawkins:
This is a little pie in the sky, but what about making the concepts more composable?

Option 1: 

When there are multiple subtypes, have the withXXX method take a fluent class as an argument:

```withNewField(SomethingFluent.class)...end()```

The method signature would look something like:

```public <T extends ThingFluent<T,  ParentFluent<A, Parent>> T withNewField(Class<T> fluent)```

Such that:
- ThingFluent is a marker interface, which is implemented by the respective fluents of the subclasses.
- Fluents would need another type parameter and a way to pass in a Supplier of that type. The default implementation of end would be to build to fluent.

Other benefits:
- no chance for a name collision
- evolves a little better, a common refactoring like renaming a subtype class would be handled automatically. 

Drawbacks:
- All fluents / builders would have an end method
- IDE autocompletion will likely be more limited

Option 2:

Aggregate the subtype methods to a single separate fluent. This would look roughly like:

```withNewField().asSomething()...end()```

This is somewhat similar to the above, but would create an intermediate fluent where each subtype had an asXXX method.

Benefits:
- Keeps the explosion of subtype methods to a single(?) place.
- Provide better autocomplete support than option 1.

Additional Drawbacks:
- Brings back name collision issues.

With either of these options we could consider adding these methods even when there is a single subtype so that users would have the option of using an api what would remain stable even if they later add more subtypes.

@manusa @iocanel WDYT?

manusa:
What about of a combination of your Option 2 (separate fluents) which will minimize the top-level fluent impl, plus what Ioannis suggested (withXxxAsZzz):
- Keeps the explosion of subtype methods to a single(?) place.
- Doesn't have the name collision issues

In any case, for me the main problem is about the name collision.

Adding additional configuration options for Buildable would also be nice. IMO being able to provide a list of interfaces/classes that should be ignored for implementation discovery would also be quite useful and would tackle the issue at hand (from the Fabric8 side).

iocanel:
I would take `Option 1` if we could eliminate the drawbacks. We could possible keep `Fluent` free of `parent` related parameters and move everything in the `Nested` class implementation which is something the users almost never directly use. 

Can it stand as a solution on its own. Possibly not. Users should not need to reference `Fluent` classes or any classes at all is it kind of defeats the purpose of `withNewXXX` as it becomes `withXXX(new XXXBuilder() ... ).

I could see that combined with something else, provided it does not add to the complexity of the internals.

shawkins:
> What about of a combination of your Option 2 (separate fluents) which will minimize the top-level fluent impl, plus what Ioannis suggested (withXxxAsZzz):

You don't need both - you can either do the top level methods, or have an intermediate fluent:

withNewSomethingAsField()....endField()
vs
withNewField().asSomething()...end()

Doing the withNewSomethingAsField requires each such field to introduce N subtype methods and N subtype nested impls where N is the number of subtypes. The other approach requires only 1 method and 1 nested per field.

> In any case, for me the main problem is about the name collision.

We still have name collision issues for anything that doesn't use a class / instance. Even without the name normalization that drops repeated terms you still have to account for package level conflicts (I believe the logic already handles this), but there will always be drawbacks and evolution issues.

> We could possible keep Fluent free of parent related parameters

I don't see how we could do that and still transition from end() back to the parent builder.

> Users should not need to reference Fluent classes or any classes at all is it kind of defeats the purpose of withNewXXX as it becomes `withXXX(new XXXBuilder() ... ).

Correct, withNewField(SomethingFluent.class)...end() is getting very close to withField(new SomethingBuilder() ...build()). It just provides a little more type hinting. However at least in eclipse that's not enough for it suggest completion targets - I'm not sure about the other IDEs. So maybe this approach isn't worth the effort.

shawkins:
@iocanel @manusa also I wasn't aware there is also already the IgnoreDescendants annotation that could have done what Marc was looking for if we stuck with the KubernetesResource.

manusa:
> @iocanel @manusa also I wasn't aware there is also already the IgnoreDescendants annotation that could have done what Marc was looking for if we stuck with the KubernetesResource.

OK, I just became aware of this too. This probably deserves some documentation in the [Builder annotation doc page](https://github.com/sundrio/sundrio/blob/main/annotations/builder/readme.md).

However, this is not exactly what I described since this configuration is based on the interface and not on the class annotated with buildable. This way it cannot be set selectively.

But yes, for the Fabric8 case this could be useful or a good alternative.

shawkins:
> However, this is not exactly what I described since this configuration is based on the interface and not on the class annotated with buildable. This way it cannot be set selectively.

It's a field level annotation.

> withNewField().asSomething()...endField()

Another drawback here is that there's no great place to define the intermediate fluent. The buildable subtypes for a type like KubernetesResource will vary by module. We also don't want to keep re-defining the same class as that will lead to package conflicts in modules. This implies per run of the logic on the first encounter of a field with a super type we'd have to create the intermediate fluent in the same package as that class, then keep referencing it when we encounter that type again.

manusa:
> It's a field level annotation.

Oh, OK, then that's the perfect fit for us.

iocanel:
I think that this can be now closed, as I just merged @shawkins PR which includes this test.

PR #483, opened 2024-09-27
Iss170

shawkins:
@iocanel @manusa consolodates the descendant logic and adds a collision check as to whether compact names can be used. This is not a major overhaul, or a breaking change, just an incremental change to handle collisions that was otherwise broken.


iocanel (approved):
LGTM

PR #487, opened 2024-12-02
Make the `generate-bom` mojo threadsafe

gastaldi:
This allows the plugin to be run without warnings when running in parallel


andreaTP:
The definition is:

> a mojo is considered threadSafe as long as there's no mutable state that is shared across different instances of the mojo -- e.g. a static reference to an object with mutable contents.

I'm, personally, not familiar with this part of the codebase.
Have you already done all the diligence?

gastaldi:
AFAIR the mojo runs only once and AFAIR I don't believe it has shared state, but I believe @iocanel knows better than me

iocanel (approved):
LGTM

PR #491, opened 2025-02-03
chore: prepare for 0.200.1

shawkins:


andreaTP (approved):
Thanks a lot for taking care of this @shawkins ! 🙏

.github/project.yml
```diff
@@ -1,4 +1,4 @@
 name: Sundrio
 release:
   current-version: 0.200.0
```
iocanel, line 3:
```suggestion
  current-version: 0.200.1
```

iocanel (changes requested):
I think that you updated the wrong property. 
To speed things up, you can apply the suggestion and merge.

.github/project.yml
shawkins: Sorry, I've pushed a commit to correct this. After reviewing the release workflow more it looks like the next-version needed corrected as well.

PR #495, opened 2025-02-27
Maven 4

iocanel:
To speed things up I copied @velo commit here and applied @gnodet suggestion


gnodet:
See https://github.com/iocanel/sundrio/pull/1
I would keep Maven 3.9.9 to build.
The problem is that Maven 4 is more strict wrt to profiles.  The `with-examples` profiles is, for an unknown reason, passed to the IT, and given it does not exist, fail the build.

PR #500, opened 2025-03-10
chore: prepare release 0.200.3

iocanel:


manusa:
@iocanel please note that the release workflow hasn't completed for a while. It seems there might be authentication issues. (Maybe an expired [Access User Token](https://oss.sonatype.org/#profile;User%20Token))

https://github.com/sundrio/sundrio/actions/workflows/release.yml

iocanel:
> @iocanel please note that the release workflow hasn't completed for a while. It seems there might be authentication issues. (Maybe an expired [Access User Token](https://oss.sonatype.org/#profile;User%20Token))
> 
> https://github.com/sundrio/sundrio/actions/workflows/release.yml

You are right and I did have to update the secrets. The workflow still failed, not sure why but at leat the release is now on central.

PR #538, opened 2025-08-29
Improve the model fluency

iocanel:
The pull request introduces auxillary DSLs that can be used for the following:

- for
- while
- if
- switch
- and more

Additionally, it refactors almost all uses of StringStatement, replacing them with the actual API use.

The pull request improves the code base in the following ways:
- Code maintainability
- Generated code readibility and correctness
- Flexibility refactorability.

## Flexibility refactorability
An issue with string concatenation is the time of the evaluation (expressions are rendered partially when we concatenate the strings). There are a few cases in our code base where this is uneeded and problematic, which lead to the use of `StingSupplier` which makes code harder to read and maintain. With the use of the API the code is rendered on demand and not on TypeDef defintion, meaning that it opens up the way for manipulation.

### DIFF
--- a/model/base/src/main/java/io/sundr/model/IfDslThenStep.java
+++ b/model/base/src/main/java/io/sundr/model/IfDslThenStep.java
@@ -31,6 +31,14 @@ public class IfDslThenStep implements Statement {
     return new If(condition, Block.wrap(statements));
   }
 
+  void setCondition(Expression condition) {
+    this.condition = condition;
+  }
+
+  void setStatements(List<Statement> statements) {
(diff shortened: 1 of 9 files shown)


iocanel:
BTW, the changes seem to work with the kubernets-client.

iocanel:
@metacosm: Thanks for having a look. It does improve the legibility and maintainability of the code.

manusa:
Did a quick pass over every changed file.
Doesn't look like it should introduce breaking changes downstream.
Looks good.

PR #542, opened 2025-09-01
Avoid conflict with Java package names (#1)

ldanesi:
When trying to use CRDS from OpenTelemetry, I encountered an issue with the generated code: Instrumentor has a field names java and some methods use FQDN beginning by java like java.util.Optional or java.util.Objects.
So I replaced FQDN by imports in the concerned Method/Class.
In addition, there were some calls using replaceAll with RegExp that were failing on Windows as File.separator was interpreted as the escape character without anything after.
I changed these calls by replace method that replace a given String by another. They do functionally the samething.

### DIFF
--- a/adapters/source/src/main/java/io/sundr/adapter/source/TypeDeclarationLookup.java
+++ b/adapters/source/src/main/java/io/sundr/adapter/source/TypeDeclarationLookup.java
@@ -21,7 +21,6 @@ import java.io.File;
 import java.io.FileInputStream;
 import java.io.IOException;
 import java.util.Optional;
-import java.util.regex.Pattern;
 
 import com.github.javaparser.ast.body.TypeDeclaration;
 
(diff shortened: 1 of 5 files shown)


iocanel:
@ldanesi thanks for the pull request. I am currently working on #538 which is pretty large and will definitely conflict with this. 
I will revisit this once #538 is merged and I'll deal with the conflicts myself, unless you volunteer.

iocanel:
@ldanesi to give you an update...
- [x] Rebased your PR to latest main.
- [x] Fixed conflicts
- [ ] Tryin to adapt your FQCN change to the new API (it needs a bit of work).

iocanel:
Couldn't push on @ldanesi branch so I pushed his commits, rebased and added some needed fixes as part of #545.
Closing this one.

PR #583, opened 2026-04-01
fix: jdk-25 nested fluent builder issue

### LINKED ISSUE #573: JDK 25: Cannot introspect nested types from JAR files - incomplete builder generation even with --add-opens
### Summary

  Sundrio cannot introspect nested type chains from JAR files on JDK 25, resulting in incomplete builder generation. This works correctly on JDK 17/21 but fails on JDK 25 due to stricter module encapsulation that cannot be bypassed even with `--add-opens` flags.

  ### Environment

  - **JDK Version**: OpenJDK 25.0.2 (Temurin)
  - **Sundrio Version**: 0.230.1
  - **Maven Version**: 3.9.x
  - **Works on**: JDK 17, JDK 21
  - **Fails on**: JDK 25

  ### The Issue

  When a `@Buildable` class in a consumer module references classes from a dependency JAR:

  1. **Without `--add-opens`**: Builders are not generated at all
  2. **With `--add-opens`**: Builders are generated but incomplete - missing nested collection builder methods

  **Example:**
  - `PolicyBinding` has `@BuildableReference(Selector.class)`
  - `Selector` contains `List<Requirement>` field
  - On JDK 17/21: Sundrio introspects `Selector` from JAR and discovers `Requirement` → generates complete builders ✅
  - On JDK 25: Sundrio reads `Selector` from JAR but CANNOT discover `Requirement` → generates incomplete builders ❌

  **Result**: Missing `addNewMatchExpression()` methods in generated `SelectorFluent` class.

  ### Reproduction

  Minimal reproducible example: https://github.com/ash-thakur-rh/sundrio-jdk25-issue

  **Steps to reproduce:**

  1. **Build on JDK 17/21** (works):
   ```bash
    export JAVA_HOME=/path/to/jdk-17
    mvn clean install
    # ✅ Build succeeds - complete builders generated
   ```

  2. Build on JDK 25 without --add-opens (fails):
  ```bash
  export JAVA_HOME=/path/to/jdk-25
  mvn clean install
  # ❌ Build fails - builders not generated
  # [ERROR] PolicyBindingBuilder not found
  ```

  3. Build on JDK 25 with --add-opens (fails differently):
  ```bash
  export JAVA_HOME=/path/to/jdk-25
  # Uncomment --add-opens compiler args in pom.xml (lines 46-75)
  mvn clean install
  # ❌ Build fails - builders generated but incomplete
  # [ERROR] cannot find symbol: method addNewMatchExpression()
  ```

  Expected Behavior

  On JDK 25, Sundrio should introspect nested types from JAR files just like it does on JDK 17/21, generating complete builders with all nested collection methods.

  Actual Behavior

  On JDK 25, even with --add-opens flags, Sundrio cannot introspect nested types from JAR files:
  - It reads the parent class (Selector) from the JAR
  - It sees the collection field (List<Requirement>)
  - It cannot introspect into the nested type (Requirement) to generate builder methods
  - Results in incomplete builders missing addNew*() methods

  Root Cause

  JDK 25's stricter module encapsulation prevents Sundrio from performing deep reflection into classes loaded from JAR files. The --add-opens flags allow basic builder generation but don't enable the nested type discovery that
  worked on JDK 17/21.

ash-thakur-rh:
Fixes #573

### DIFF
--- a/annotations/builder/src/main/java/io/sundr/builder/internal/utils/BuilderUtils.java
+++ b/annotations/builder/src/main/java/io/sundr/builder/internal/utils/BuilderUtils.java
@@ -47,8 +47,10 @@ import java.util.stream.Collectors;
 import java.util.stream.Stream;
 
 import javax.lang.model.element.Element;
+import javax.lang.model.element.ElementKind;
 import javax.lang.model.element.TypeElement;
 import javax.lang.model.type.MirroredTypeException;
+import javax.lang.model.util.Elements;
 
 import io.sundr.adapter.api.AdapterContext;
 import io.sundr.adapter.api.Adapters;
@@ -151,7 +153,28 @@ public class BuilderUtils {
   }
 
   public static boolean isBuildable(TypeRef typeRef) {
-    return isRegisteredAsBuildable(typeRef) && canBeBuilt(typeRef);
+    if (isRegisteredAsBuildable(typeRef) && canBeBuilt(typeRef)) {
+      return true;
+    }
+    if (!(typeRef instanceof ClassRef)) {
+      return false;
+    }
+    // Fallback for cross-module @BuildableReference: check if a sundrio-generated *Builder exists
+    // on the APT classpath. @Buildable has SOURCE retention so compiled JARs don't carry it;
+    // using getTypeElement() is the only JDK-version-agnostic way to detect a pre-compiled builder.
+    // We verify it's a sundrio builder (not e.g. java.lang.StringBuilder or ProcessBuilder) by
+    // checking it declares a build() method.
+    String builderFQCN = ((ClassRef) typeRef).getFullyQualifiedName() + "Builder";
+    Elements elements = BuilderContextManager.getContext().getElements();
+    if (elements == null) {
+      return false;
+    }
+    TypeElement builderElement = elements.getTypeElement(builderFQCN);
+    if (builderElement == null) {
+      return false;
+    }
+    return builderElement.getEnclosedElements().stream()
+        .anyMatch(e -> e.getKind() == ElementKind.METHOD && e.getSimpleName().contentEquals("build"));
   }
 
   public static boolean isRegisteredAsBuildable(TypeDef typeDef) {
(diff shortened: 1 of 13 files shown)


annotations/builder/src/main/java/io/sundr/builder/internal/utils/BuilderUtils.java
```diff
@@ -151,7 +153,28 @@ public static boolean isRegisteredAsBuildable(TypeRef typeRef) {
   }
 
   public static boolean isBuildable(TypeRef typeRef) {
-    return isRegisteredAsBuildable(typeRef) && canBeBuilt(typeRef);
+    if (isRegisteredAsBuildable(typeRef) && canBeBuilt(typeRef)) {
```
iocanel: For a type to be considered buildable we need:
- to be registered.
- to have the means to build it.

In other words we want it to be either annotated as `Buildable` in the current module, or define it as `BuildableReference` in another module. This covers the `registered` part.

For something to be able to be built, we make additional checks.

The suggested change opens it to cases where a type `hasBuilder`. 

If I were to accept this change I would need to see a hasBuilder method and some kind of logic like: `retrun (isRegistered(type) && canBeBuilt(type)) || hasBuilder(type)`. But first I am going to need some conviencing on why we need something like that and how its related to java 25.

ash-thakur-rh: Thank you for the detailed feedback. I've restructured the change as suggested:
- Extracted a dedicated hasBuilder(ClassRef) method that checks first in the definition repository (for same-module types), then falls back to the APT Elements API for cross-module types.
- Updated isBuildable(TypeRef) to use the logic you proposed: `isRegisteredAsBuildable(type) && canBeBuilt(type)) || hasBuilder(type)`

### Why I think this is needed and how it relates to cross-module / JDK 25:
- `@Buildable` has SOURCE retention, so it is not present in compiled JARs. When a consumer module uses `@BuildableReference` to reference a type from a library that was compiled separately, the annotation processor has no way to detect that the type is buildable via the repository, it was never registered in this compilation unit.
- This scenario is reliably reproducible when annotationProcessorPaths is used (as is common with JDK 21+/25 where processor isolation is stricter), because the processor classpath is separate from the compile classpath. The only JDK-version-agnostic way to detect a pre-compiled builder is `Elements.getTypeElement()`, which is what hasBuilder uses as a fallback.
- Without this, the processor silently omits generated methods like addNewMatchExpression() from the fluent builder, no error, just missing API.
- I've also added an integration test (`cross-module-buildable-ref`) that reproduces the failure and verifies the fix.

iocanel (approved):
LGTM. I am still not super happy with `isBuildable`, `canBeBuilt`  and `hasBuilder`, as the separation is not as clear as I would like to. Still, this may be a bit of the scope of this merge request, so I'll accept the merge request as is. @ash-thakur-rh feel free to create a follow up merge request and suggest more clean boundaries between these methods now that you have been exposed to it.

PR #600, opened 2026-09-12
Reduce common compiler warnings in generated code

MikeEdgar:
This change reduces the number of common compiler warnings in generated code. These are only the ones I see in my project, and it is not meant to address every possible instance.

- Only add `unchecked` warning suppression for fluent classes with fields
- Add `@Generated` annotation to classes with sundr.io as generator and current timestamp (not a warning, just nice to have)
- Tighten the definition of local variables in builder constructors and `build` method to avoid unused `java.lang.Object` leaking into class imports
- Suppress `rawtypes` on local `List` variables read from `_visitables`
- Add wildcards to casts in `equals` methods
- Avoid creating unused local `that` variable when there are no fields to compare
- Add diamond operator to lazy-initialized collections/maps


MikeEdgar:
@iocanel please let me know if you're open to merging this or similar. The intent is to reduce the warning "noise" users may see, without changing any functionality.

annotations/builder/src/main/java/io/sundr/builder/internal/functions/ClazzAs.java
```diff
@@ -755,4 +759,45 @@ private static Predicate<Field> isFieldApplicable(RichTypeDef item, boolean excl
...
+  }
+
+  /**
+   * Create a Generated annotation with the current date and time.
+   */
+  private static AnnotationRef getGenerated() {
+    return new AnnotationRefBuilder()
+        .withClassRef(ClassRef.forName(javax.annotation.processing.Generated.class.getCanonicalName()))
+        .addToParameters("value", "io.sundr:builder-annotations")
+        .addToParameters("date", OffsetDateTime.now(Clock.systemDefaultZone()).truncatedTo(ChronoUnit.MILLIS).toString())
```
iocanel: I don't really like having a timestamp in the source. Is there a reason to have it ?

MikeEdgar: No, not really. I'll take that out. I thought maybe it was required, but there is a blank default on `@Generated`.

MikeEdgar: Removed `date`. I can squash this before we're done.

iocanel (approved):
LGTM

Issue #599, opened 2026-09-11
Builder generation bug: @Accessors prefix not respected in copyInstance() for underscore-prefixed fields


ash-thakur-rh:
## Description

Builder generation does not properly handle fields with underscore prefixes when `@Accessors(prefix = {"_", ""})` is used. The generated `copyInstance()` method omits the `withXxx()` call for such fields, causing data loss when initializing builders from existing instances.

## Affected Versions

- ✅ **Working**: Sundrio 0.230.2 and earlier
- ❌ **Broken**: Sundrio 0.240.1, 0.300.0, 0.400.0 (and likely all versions after 0.230.2)

## Root Cause

Commit `c1303fe` ("intermediate: separate properties into fields, arguments and local variables") refactored the `Property` class into separate `Field`, `Argument`, and `LocalVariable` types. This refactoring broke the logic that handles `@Accessors` prefix stripping when generating the `copyInstance()` method.

**Key commit**: https://github.com/sundrio/sundrio/commit/c1303fe

**Files affected**: `BuilderUtils.java` (142 lines changed) and related builder generation code.

### 1. Create a class with underscore-prefixed field:

```java
@Buildable
public class ListMeta {
    @JsonProperty("continue")
    private String _continue;  // Underscore required (continue is a keyword)
    
    @JsonProperty("resourceVersion")
    private String resourceVersion;

    @Accessors(prefix = {"_", ""})
    
    public String getContinue() {
        return _continue;
    }
    
    public void setContinue(String _continue) {
        this._continue = _continue;
    }
    
    public String getResourceVersion() {
        return resourceVersion;
    }
    
    public void setResourceVersion(String resourceVersion) {
        this.resourceVersion = resourceVersion;
    }
}
```

### 3. Observe generated `ListMetaFluent.copyInstance()`:

**Expected (Sundrio 0.230.2)**:
```java
protected void copyInstance(ListMeta instance) {
    instance = instance != null ? instance : new ListMeta();
    if (instance != null) {
        this.withContinue(instance.getContinue());  // ✅ Present
        this.withResourceVersion(instance.getResourceVersion());
    }
}
```

**Actual (Sundrio 0.240.1+)**:
```java
protected void copyInstance(ListMeta instance) {
    instance = instance != null ? instance : new ListMeta();
    if (instance != null) {
        // ❌ Missing: this.withContinue(instance.getContinue());
        this.withResourceVersion(instance.getResourceVersion());
    }
}
```

## Test Case

```java
[...]

PR #602, opened 2026-09-22
fix: 0.400.x regressions breaking the Kubernetes Client builders

manusa:
Updating the Fabric8 Kubernetes Client from 0.230.2 to 0.400.x (fabric8io/kubernetes-client#8138) surfaced two regressions in the builder generator, both present in 0.400.0 and 0.400.1. This PR fixes them, adds tests reproducing each one, and fixes the downstream check job so it can catch this kind of problem again.

## Escaped keyword fields are dropped when copying an instance

Fields named after Java keywords are usually escaped with a leading underscore while their accessors aren't, e.g. `_continue` with `getContinue()`/`setContinue()` (typical of code generated from OpenAPI or JSON schemas).

The `Getter.find` rewrite that came with the `Property` → `Field` refactoring matches `getContinue` by its `Continue` suffix, but then checks the length against the raw field name: `"get".length() + "_continue".length()` is 12, while `getContinue` is 11. The getter is rejected and `copyInstance` silently skips the field, so `new FooBuilder(foo)`, `edit()` and nested `withNewX()...endX()` lose the value without any error. In the Kubernetes Client that's about 40 model fields (`ListMeta._continue`, `JSONSchemaProps._default`, Tekton's `_finally`, `_int`, `_import`, `_interface`...).

The fix compares whole method names, trying the names a getter may use closest to the field name first: the bean-style name (`get_Id()`, `getName()`), then the raw name for names not starting with a letter (`get_1h()`, `get$ref()`), then the normalized one (`getContinue()`). So besides `_continue`:

- `_1h` / `get_1h()` (e.g. Lombok) keeps working, 0.400.x accepted it through the length check.
- When both `getId()` and `get_id()` exist, `_id` gets `get_id()` regardless of declaration order.
- A bare `get()` is never taken as the getter of a field without alphanumeric characters.

Setters are left alone: `set_1h()` isn't recognized as the setter of `_1h`, but that's unchanged since 0.230.2.

Tests: `GetterTest` (escaped keyword, `is` prefix, `$ref`, `_1h`, another getter with the same suffix, declaration order, bare `get()`), plus `EscapedFields` in `tests/buildable-fields` going through the processor (constructor arguments and setter).

## Processor crash when a hand-written builder extends a fluent generated in the same round

Since ca0c5518 ("make hasBuilder more accurate"), `BuilderUtils.hasBuilder` looks up `<Type>Builder` and checks it's assignable to the builder interface, which walks the builder's whole type hierarchy. If that hierarchy contains a type that is generated in the current round, javac returns a type variable with a `null` upper bound and `TypeParameterElement.getBounds()` throws:

```
io.sundr.SundrException: Failed to get bounds of type: T.This may be caused due to the compiler not being able to resolve a symbol. Please check for unresolved symbols! Cannot invoke "com.sun.tools.javac.code.Type.isCompound()" because "bound" is null
```

The Kubernetes Client hits it in `openshift-client-api`: `@Buildable SundrioOpenShiftConfig extends OpenShiftConfig`, and the hand-written `OpenShiftConfigBuilder` extends `OpenShiftConfigFluent`, which extends the generated `SundrioOpenShiftConfigFluent`. While generating `SundrioOpenShiftConfig`, `hasBuilder(OpenShiftConfig)` finds `OpenShiftConfigBuilder` and walks into `SundrioOpenShiftConfigFluent` before it exists.

The fix makes `TypePrameterElementToTypeParamDef` return the type parameter without bounds when its bound can't be resolved, which is what the comment in the existing catch block described. The builder is still recognized through its `VisitableBuilder` interface, and the generated `SundrioOpenShiftConfigFluent` has the same API as with 0.230.2. I also added the declaring element to the remaining error message, since "type T" alone was hard to track down.

Handling it in `hasBuilder` instead (e.g. checking the implemented interfaces before the superclass) would also work for this case, but fixing the adapter covers any other caller that adapts a partially resolved type.

Test: a new `tests/custom-builder` module reproducing that layout, which doesn't compile without the fix.

## Downstream check

The downstream check has been failing in the Sundrio build step itself: it runs on JDK 11, and the tests now use JUnit 6 (class file version 61). It also couldn't have caught the first issue, because the Kubernetes Client commits the generated builders and only regenerates them with `-Pgenerate`.

- JDK 17 and the same action versions as `build.yml`
- Regenerate the Kubernetes Client builders (`make openapi-generate-java-classes`, Maven only) and print a `git diff --stat` of the result before building it. Tests are skipped while regenerating, the build runs them.
- Fail early if the Sundrio version can't be resolved, instead of passing an empty `sundrio.version`

With the `downstream-check` label it [passed on this PR](https://github.com/sundrio/sundrio/actions/runs/35699467679): 10,161 builder files regenerated, then the full Kubernetes Client build and tests.

## Verification

- `mvn clean install -Pwith-examples -Pwith-tests` passes. The new tests fail on current `main` (`EscapedFieldsTest`: `expected: <token> but was: <null>`; `custom-builder`: the crash above) and pass with the fixes.
- Kubernetes Client built against these changes: all model builders regenerated, and the full build and test suite pass, except three tests that also fail locally on its main branch. The regenerated `kubernetes-model-core` has the same public API as the one generated with 0.230.2.

Could we get a 0.400.2 release with these fixes? We'd like to include the Sundrio update in Kubernetes Client 8.0.0.


iocanel (approved):

manusa:
This is ready now.
@iocanel, thx for reviewing. If you merge and release we can try to include the Sundrio bump as part of the Kubernetes Client 8.0.0 release (probably tomorrow or Thursday).

ash-thakur-rh:
This PR fixes: https://github.com/sundrio/sundrio/issues/599

